PR #2, opened 2026-03-10
feat(jira): support custom fields in create_issue

decko:
## Summary

- Extract type conversion logic from `set_custom_field` into a shared `resolve_field_value()` helper in `helpers.py`
- Add `custom_fields` parameter to `jira_create_issue` for setting custom fields at creation time
- Refactor `set_custom_field` to use the shared helper, eliminating duplicate code

### Usage

```json
{
  "project": "PROJ",
  "issue_type": "Task",
  "summary": "My new task",
  "custom_fields": [
    {"field_id": "customfield_12345", "value": "rhel-10", "field_type": "version"},
    {"field_id": "customfield_99999", "value": 5.0, "field_type": "number"},
    {"field_id": "customfield_11111", "value": "High", "field_type": "select"}
  ]
}
```

Supported `field_type` values: `auto`, `text`, `number`, `select`, `multi-select`, `version`, `user`

## Test plan

- [ ] Verify `jira_create_issue` with `custom_fields` in dry_run mode returns correct field values
- [ ] Verify `jira_create_issue` with `custom_fields` creates issue with custom fields set
- [ ] Verify `jira_set_custom_field` still works correctly with the refactored helper
- [ ] Verify `create_issue` without `custom_fields` is unchanged (backward-compatible)

### DIFF
--- a/plugins/jira/tools_write.py
+++ b/plugins/jira/tools_write.py
@@ -45,9 +45,13 @@ def create_issue(params):
         fields["components"] = normalize_components(comp_list)
 
     for cf in params.get("custom_fields", []):
-        fields[cf["field_id"]], _ = resolve_field_value(
-            cf["value"], cf.get("field_type", "auto"), is_cloud=handler.is_cloud
-        )
+        cf_id = cf["field_id"]
+        if not cf_id.startswith("customfield_"):
+            raise ValueError(
+                f"Invalid custom field ID: '{cf_id}' (must start with 'customfield_'). "
+                "Use the standard parameters for built-in fields."
+            )
+        fields[cf_id], _ = resolve_field_value(cf["value"], cf.get("field_type", "auto"), is_cloud=handler.is_cloud)
 
     if dry_run:
         return {"dry_run": True, "action": "jira_create_issue", "fields": fields}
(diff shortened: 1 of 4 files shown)


plugins/jira/tools_write.py
```diff
@@ -45,7 +45,7 @@ def create_issue(params):
         fields["components"] = normalize_components(comp_list)
 
     for cf in params.get("custom_fields", []):
-        fields[cf["field_id"]] = resolve_field_value(
+        fields[cf["field_id"]], _ = resolve_field_value(
             cf["value"], cf.get("field_type", "auto"), is_cloud=handler.is_cloud
```
sergio-correia: Let's make sure `field_id` is actually a custom field, to prevent the case we may override standard field.s. You can probably check if the field starts with`customfield_`, and reject any attempt to override standard fields with an error message.

sergio-correia (changes requested):

plugins/jira/tools_write.py
decko: Done sir

sergio-correia (approved):
Thanks, LGTM

PR #3, opened 2026-03-11
feat: add wtmcpctl OAuth management tool

rjeffman:
Created wtmcpctl, a command-line utility for managing OAuth authentication for Google plugins:
- Added wtmcpctl tool with oauth list and oauth auth commands
- List command shows authentication status for all Google plugins
- Auth command implements OAuth2 consent flow with out-of-band redirect URI (urn:ietf:wg:oauth:2.0:oob)
- Upgraded Google Docs plugin to full read/write/create scope (documents instead of documents.readonly)
- Reorganized cmd/ directory to support multiple binaries (wtmcp and wtmcpctl)
- Updated Makefile to build both binaries
- Added comprehensive documentation in docs/wtmcpctl.md
- Updated README.md with OAuth plugin management section
- Fixed .gitignore to use absolute paths for binaries

Prompts:
- Create a new tool wtmcpctl which accepts commands and maybe, subcommands. For command oauth it will operate on any plugin that has the attribute auth: oauth. The subcommand list will list the name of each plugin with auth: oauth and a check mark or an X icon, to show that the plugin credentials file token-*.json is available. The subcommand auth will use Google OAuth Consent link on the terminal, so user can open it in the browser, wait for an authentication code, and with the code, create the token-*.json file. Write this tool using Go, and use the same libraries used to manage Google plugins in wtmcp.
- Use the url 'urn:ietf:wg:oauth:2.0:oob' as redirect_uri in the consent link and ask for read, write and create scopes for documents.
- Use version '0.1.0', don't use 'BuildDate'


cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,330 @@
...
+	TokenFile string
+	Scopes    []string
+}
+
+// googlePlugins is the registry of all Google OAuth plugins.
+var googlePlugins = []GooglePlugin{
+	{
+		Name:      "google-docs",
+		TokenFile: "token-docs.json",
+		Scopes:    []string{"https://www.googleapis.com/auth/documents"},
```
sergio-correia: This should not be here yet.

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	cloud.google.com/go/auth/oauth2adapt v0.2.7 // indirect
 	cloud.google.com/go/compute/metadata v0.6.0 // indirect
 	github.com/JohannesKaufmann/dom v0.2.0 // indirect
+	github.com/LeGambiArt/oauth2flow v0.2.0 // indirect
```
sergio-correia: This is not indirect, as it is directly imported in `cmd/wtmcpctl/main.go` -- running `go mod tidy` would likely fix this.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	fmt.Println()
+	fmt.Printf("✓ Successfully authenticated %s\n", plugin.Name)
+	fmt.Printf("Token saved to: %s\n", tokenPath)
+	return nil
+}
+
+// tokenJSON matches the on-disk format for OAuth2 tokens.
+type tokenJSON struct {
```
sergio-correia: These seem exactly the same as the ones in `internal/google/auth.go` -- you should export it in `auth.go` and reuse here.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,330 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/LeGambiArt/oauth2flow"
+	"golang.org/x/oauth2"
+
+	googleauth "github.com/LeGambiArt/wtmcp/internal/google"
+)
+
+const Version = "0.1.0"
```
sergio-correia: This being a const, it won't be able to be overidden by ldflags and it will always be `"0.1.0"`. In the `Makefile` we have `-X main.Version=$(VERSION)`; make this a `var`, like we have in `cmd/wtmcp/main.go`.

sergio-correia (changes requested):
Thanks, I like the idea here. A couple more issues we should address before merging:
-  `wtmcpctl` should accept a `--workdir` flag (or at least `--credentials-dir`). `wtmcp` supports `--workdir` which changes where credentials live. Without this, users with a custom `workdir` will have the two tools looking in different places.
- plugin names passed to `oauth auth` should be validated upfront before starting any auth flows. Right now a typo like `google-doc` will silently proceed and only fail mid-flow. With `oauth auth google-drive google-doc google-calendar`, it would auth `google-drive` successfully, then crash on the typo, leaving the user in some kind of partial state.

go.mod
rjeffman: Fixed.

cmd/wtmcpctl/main.go
rjeffman: Fixed.

cmd/wtmcpctl/main.go
rjeffman: Fixed.

cmd/wtmcpctl/main.go
rjeffman: This was added when I mistakenly allowed claude-code to accept all edits. I knew it would bite me. :-D

Change the code to search for plugins that require oauth authentication and extract data from the plugin itself. This requires some changes to the plugins, wich I think is ok.

Let me know if you want the single huge commit to be splitted into different ones.
The changes are all related anyway, and having a single commit will not cause any "ordering issue".

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+// PluginManifest represents the structure of plugin.yaml
+type PluginManifest struct {
+	Name     string `yaml:"name"`
+	Services struct {
+		Auth struct {
+			Type            string   `yaml:"type"`
+			TokenFile       string   `yaml:"token_file"`
+			CredentialsFile string   `yaml:"credentials_file"`
```
sergio-correia: `credentials_file` does not seem to be used. We read it here but further on we hardcode it in line 351. Either use it from the manifest file insted of harcdoding or remove the field here.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+	if plugin == nil {
+		return fmt.Errorf("unknown plugin: %s", pluginName)
+	}
+
+	credDir := googleauth.CredentialsDir()
+	if credDir == "" {
+		return fmt.Errorf("cannot determine credentials directory")
+	}
+
+	clientCredsPath := filepath.Join(credDir, "client-credentials.json")
```
sergio-correia: This is where the credentials file is hardcoded, as per the previous comment.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+
+// OAuthPlugin represents an OAuth plugin configuration.
+type OAuthPlugin struct {
+	Name      string
+	TokenFile string
+	Scopes    []string
+}
+
+// PluginManifest represents the structure of plugin.yaml
+type PluginManifest struct {
```
sergio-correia: This is redundant, as we are redeclaring what already exists in `internal/plugin/manifest.go` as `Manifest` with `AuthServiceConfig`. Let's import and reuse it.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+	"golang.org/x/oauth2"
+	"gopkg.in/yaml.v3"
+
+	googleauth "github.com/LeGambiArt/wtmcp/internal/google"
+)
+
+// Version is set via ldflags at build time.
+var (
+	Version   = "dev"
+	BuildDate = "unknown"
```
sergio-correia, line 13: `BuildDate` is declared here but never printed. We even set it (`Makefile`), but when printing the version in line 63, we don't print the date. Either use or remove it.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+	if _, err := os.Stat(pluginsDir); err != nil {
+		// Try current directory
+		if cwd, err := os.Getwd(); err == nil {
+			pluginsDir = filepath.Join(cwd, "plugins")
+		}
+	}
+
+	entries, err := os.ReadDir(pluginsDir)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read plugins directory %s: %w", pluginsDir, err)
```
sergio-correia: This discovery path is a bit fragile: we use `os.Executable()` + some relative path heuristic. It won't work after `go install`  (binary in  `$GOPATH/bin`). The core's `plugin.Discover()` solves this properly, let's share that logic.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+	default:
+		fmt.Fprintf(os.Stderr, "Unknown oauth subcommand: %s\n", subcommand)
+		printUsage()
+		os.Exit(1)
+	}
+}
+
+func printUsage() {
+	plugins, err := discoverOAuthPlugins()
```
sergio-correia: Filesystem I/O in help output is unexpected; if discovery fails, help text gets warnings. Help should be static.

docs/wtmcpctl.md
```diff
@@ -0,0 +1,241 @@
...
+- ✓ - Plugin is authenticated with a valid token
+- ! - Plugin has a token but it may be invalid or expired
+- ✗ - Plugin is not authenticated
+
+**Example:**
+```
+$ wtmcpctl oauth list
+OAuth Plugin Status:
+
+  ✓  google-docs           authenticated (valid)
```
sergio-correia: Oops, we forgot this from the initial review. Lots of other mentions in this README also.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+		for _, p := range plugins {
+			fmt.Printf("  - %s\n", p.Name)
+		}
+	}
+}
+
+// discoverOAuthPlugins scans the plugins directory and builds a list of OAuth plugins.
+func discoverOAuthPlugins() ([]OAuthPlugin, error) {
+	// Get the plugins directory relative to the binary
+	execPath, err := os.Executable()
```
sergio-correia: This discovery path is a bit fragile: we use `os.Executable()` + some relative path heuristic. It won't work after go install (binary in `$GOPATH/bin`). The core's `plugin.Discover()` solves this properly, let's share that logic.

cmd/wtmcpctl/main.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	fmt.Println()
+	fmt.Printf("✓ Successfully authenticated %s\n", plugin.Name)
+	fmt.Printf("Token saved to: %s\n", tokenPath)
+	return nil
+}
+
+// loadToken loads an OAuth2 token from a JSON file.
+func loadToken(path string) (*oauth2.Token, error) {
```
sergio-correia: Export `LoadToken` from `internal/google/auth.go` and delete this copy here.

sergio-correia (changes requested):
I mentioned in the previous review, but didn't see it addressed yet: wtmcpctl should accept a `--workdir` flag (or at least `--credentials-dir`). wtmcp supports `--workdir` which changes where credentials live. Without this, users with a custom workdir will have the two tools looking in different places.

cmd/wtmcpctl/main.go
rjeffman: Unless, I forgot something, fixed.

cmd/wtmcpctl/main.go
rjeffman: should be fixed.

cmd/wtmcpctl/main.go
rjeffman: Fixed.

cmd/wtmcpctl/main.go
rjeffman: Fixed.

docs/wtmcpctl.md
rjeffman: Shouldn't it be there when listing the oauth plugins?

cmd/wtmcpctl/main.go
rjeffman: Fixed.

cmd/wtmcpctl/main.go
rjeffman: Fixed.

docs/wtmcpctl.md
rjeffman: Finally got it. :-D
I do have google-docs around... :-D

rjeffman: Fixed.

rjeffman:
Added `--workdir` and it also respect `GOOGLE_CREDENTIALS_DIR`.

Makefile
```diff
@@ -4,20 +4,34 @@ VERSION ?= $(shell cat VERSION 2>/dev/null || echo "dev")
 BUILD_DATE := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 LDFLAGS := -s -w -X main.Version=$(VERSION) -X main.BuildDate=$(BUILD_DATE)
 
+# Google plugin handlers
+GOOGLE_PLUGIN_HANDLERS := $(wildcard plugins/google-*/handler)
```
sergio-correia: This is not used anywhere. Can we remove it?

Makefile
sergio-correia: We need to add `google-plugins` here, as it is a grouping target (doesn't produce a file).

Makefile
```diff
@@ -4,20 +4,34 @@ VERSION ?= $(shell cat VERSION 2>/dev/null || echo "dev")
...
+# Build wtmcpctl binary
+wtmcpctl: $(shell find cmd/wtmcpctl -name '*.go') $(shell find internal -name '*.go' 2>/dev/null)
+	@echo "Building wtmcpctl..."
+	go build -ldflags "$(LDFLAGS)" -o wtmcpctl ./cmd/wtmcpctl
+
+# Build Google plugin handlers
+google-plugins: $(GOOGLE_PLUGIN_TARGETS)
+
+# Pattern rule for building plugin handlers
+plugins/google-%/handler: $(shell find plugins/google-$* -name '*.go' 2>/dev/null)
```
sergio-correia: This is expanded at `Makefile` parse file, not when the pattern rule matches. At parse time, `$*` is empty, so this `find` searches `plugins/google-` which does not exist. The `2>/dev/null` hides the error. The handler targets end up with no prerequisites. If we make the plugin targets `.PHONY` (re. the comment in the first line), they always rebuild.

sergio-correia (commented):
Thanks, now it looks good, with a few remaining issues on the `Makefile`.

Makefile
rjeffman: fixed.

Makefile
rjeffman: fixed.

Makefile
rjeffman: fixed. (kind of)

sergio-correia (approved):
LGTM, thanks!

PR #4, opened 2026-03-11
Google Doc Plugin

rjeffman:
Initial implementation for the Google Docs Plugin.

It is possible to:
- read/summarize documents given their URL (I didn't tested with Google Drive search)
- create a document, and the URL is returned/printed
- modify a document by providing text (it even tries to format markdown text)

I suggest that minor issues are handled as new issues and don't block this PR. If issues are no minor, they should be fixed before merging.


plugins/google-docs/tools.go
```diff
@@ -0,0 +1,865 @@
...
+	doc, err := docsSvc.Documents.Get(docID).Do()
+	if err != nil {
+		return nil, fmt.Errorf("get document: %w", err)
+	}
+
+	// Determine insertion index
+	insertIndex := p.InsertIndex
+	if p.AppendToEnd || insertIndex == 0 {
+		// Find the end of the document (before the final newline)
+		insertIndex = doc.Body.Content[len(doc.Body.Content)-1].EndIndex - 1
```
sergio-correia, line 1179: If the API returns a document with a `nil` `Body` or empty `Content` slice, this will panic and crash the plugin. Let's take a safer approach like we do in `extractText()`. Probably something like this would do:
```
if doc.Body == nil || len(doc.Body.Content) == 0 {
    return nil, fmt.Errorf("document %s has no content body", docID)
}
```

plugins/google-docs/tools.go
```diff
@@ -0,0 +1,865 @@
...
+type writeTextParams struct {
+	DocumentIDOrURL string `json:"document_id_or_url"`
+	Text            string `json:"text"`
+	IsMarkdown      bool   `json:"is_markdown"`
+	AppendToEnd     bool   `json:"append_to_end"`
+	InsertIndex     int64  `json:"insert_index"`
+}
+
+func toolWriteText(params, _ json.RawMessage) (any, error) {
+	var p writeTextParams
```
sergio-correia: Here we may have a subtle bug: `var p writeTextParams` initializes all bools to `false`, but `plugin.yaml` declares `is_markdown` and `append_to_end` with `default: true`. When the LLM omits these fields (trusting the advertised default), the tool silently gets `false` instead of `true`,  producing plain text at index 0 instead of formatted appends.

Initialize with the intended defaults before unmarshalling, like the Calendar plugin does, e.g.  `p := writeTextParams{IsMarkdown: true, AppendToEnd: true}`.

Same issue applies to `summarizeDocumentParams`, where `include_structure` should default to `true`.

plugins/google-docs/tools.go
```diff
@@ -0,0 +1,865 @@
...
+					text:   text[pos+1 : endPos],
+					italic: true,
+				})
+				pos = endPos + 1
+				continue
+			}
+		}
+
+		// Plain character
+		segments = append(segments, markdownSegment{text: text[pos : pos+1]})
```
sergio-correia, line 815: This fallback appends a single-character segment for every plain character. For a 1000-character paragraph with no formatting, this produces 1000 `markdownSegment` entries, each generating its own `InsertText` API request in `convertMarkdownToRequests`. This will likely be very slow and hit some kind of API limits.

A possible solution here could be accumulating consecutive plain characters, or perhaps  merging adjacent unformatted segments before retturning, after the loop.

sergio-correia (changes requested):

plugins/google-docs/tools.go
rjeffman: I believe this is fixed.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: yep... this was really, REALLY, bad...

Fixed. In the test I did, it  dropped from 327 to 64.

rjeffman:
Issues were fixed and Markdown conversion was improved. Text looks good now.

There is support for most basic Markdown formatting (bold, italic, hyperlink, H1-H6, adding date objects (`@date()`), person object (`@(email)`).

Tables, images and some other artifacts are not yet supported.

rjeffman:
@sergio-correia finally the review requests and Markdown rendering are fixed!

sergio-correia (approved):
LGTM, thanks!

PR #5, opened 2026-03-16
feat: add plugins.disabled config option

thedeveloper-ctl:
## Summary

- Add plugins.disabled list to config.yaml so plugins can be disabled without modifying plugin source files
- Plugins listed in plugins.disabled are skipped during discovery and do not register any tools
- Fully backwards compatible — empty/missing list means all plugins load as before

## Usage

plugins:
  disabled:
    - testing-farm
    - gitlab

## Changes

- internal/config/config.go — add Disabled []string field to PluginsConfig
- internal/plugin/manager.go — check disabled list during Discover()
- Tests for both config loading and plugin discovery skipping


sergio-correia (approved):
Thanks, LGTM. I will add support in `wtmcpctl` for enabling/disabling the plugins.

PR #8, opened 2026-03-19
chore: update .gitignore for build and runtime artifacts

thedeveloper-ctl:
## Summary

- Generalize plugin binary pattern from `plugins/google-*/handler` to `plugins/*/handler` — covers all plugin compiled binaries (gitlab, jira, future plugins), not just Google ones
- Add `uv.lock` (generated by `uv run pytest`)
- Add `control/` (MCP runtime control directory with commands/results)

## Test plan

- [x] Verified no tracked files were removed (`git diff --name-status HEAD` shows only `.gitignore` modified)
- [x] All previously untracked build artifacts now properly ignored

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # Binaries
 /wtmcp
 /wtmcpctl
-plugins/google-*/handler
+plugins/*/handler
 
 # Test/coverage
 coverage.out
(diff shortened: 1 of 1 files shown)


sergio-correia (approved):
LGTM

PR #10, opened 2026-03-21
Add wizard to configure Google Credentials.

rjeffman:
This PR adds a wizard to guide users in configuring the credentials required for the Google plugins.

It adds a new `oauth` sub-command `credentials` and leave room for future support of other OAuth services (e.g. Github).

### DIFF
--- a/cmd/wtmcpctl/credentials.go
+++ b/cmd/wtmcpctl/credentials.go
@@ -1,0 +1,181 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+// credentialService describes a supported service for credential setup.
+type credentialService struct {
+	Name        string
+	Description string
+	SetupFunc   func() error
+}
+
+var supportedCredentialServices = make(map[string]credentialService)
+
+// registerCredentialService registers a service for credential setup.
+// This is called from init() functions in service-specific files (e.g., credentials_google.go).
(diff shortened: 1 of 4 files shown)


cmd/wtmcpctl/credentials.go
```diff
@@ -0,0 +1,176 @@
...
+		return fmt.Errorf("cannot create destination file: %w", err)
+	}
+	defer dstFile.Close() //nolint:errcheck // errors handled by explicit Close or Sync below
+
+	// Copy contents
+	if _, err := io.Copy(dstFile, srcFile); err != nil {
+		return fmt.Errorf("cannot copy file: %w", err)
+	}
+
+	return nil
```
sergio-correia, line 180: We have an issue here - `copyFile()` silently discards write errors, which can lead to data corruption without user notification.

Currently, the function defers `dstFile.Close()` and ignores its error (line 168 has a `nolint` comment claiming "errors handled by explicit Close or Sync below", but no such explicit call exists). This means:
- if the write to disk fails during Close() due to disk full, NFS errors, I/O errors, or quota exceeded, the function returns nil
- the user believes the credential file was copied successfully, but could be truncated, empty or corrupted, which will cause later authentication to fail with confusing errors.

We should explicitly close the destination file and check the error before returning. The `defer dstFile.Close()` on line 168 can remain as a safety net (closing an already-closed file is harmless in Go), but update the `nolint` comment to reflect the actual pattern.

rjeffman: Feels like learning to program again. :-)

Hopefully fixed.

sergio-correia (approved):
LGTM, thanks

PR #12, opened 2026-03-24
feat: add parent param to jira_create_issue

decko:
## Summary

- Adds an optional `parent` parameter to `jira_create_issue`, allowing child issues to be created with their parent set in a single call
- The parent key is validated via `validate_issue_key` before being included in the API payload
- Eliminates the need to call `jira_set_parent` separately after issue creation

## Test plan

- [x] `test_parent_dry_run` — parent field appears in dry-run preview
- [x] `test_parent_create_success` — creation works with parent set
- [x] `test_parent_invalid_key_raises` — invalid parent keys are rejected
- [x] All existing tests still pass

### DIFF
--- a/plugins/jira/plugin.yaml
+++ b/plugins/jira/plugin.yaml
@@ -170,6 +170,9 @@ tools:
         items:
           type: string
         description: "List of labels"
+      parent:
+        type: string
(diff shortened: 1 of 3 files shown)


sergio-correia (approved):
LGTM

PR #15, opened 2026-03-25
chore: gitignore compiled binaries, runtime state, and uv.lock

thedeveloper-ctl:
## Summary
- Add `plugins/gitlab/handler` (compiled Go binary) to `.gitignore`, matching the existing `plugins/google-*/handler` pattern
- Add `control/` (runtime state directory with commands/results) to `.gitignore`
- Add `uv.lock` (auto-generated Python UV lock file) to `.gitignore`

These files are local build artifacts and runtime state that should not be version-controlled.

## Test plan
- [x] Verify `git status` no longer shows `control/`, `plugins/gitlab/handler`, and `uv.lock` as untracked

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /wtmcp
 /wtmcpctl
 plugins/google-*/handler
+plugins/gitlab/handler
 
 # Test/coverage
 coverage.out
@@ -28,7 +29,13 @@ __pycache__/
 # Claude Code
 CLAUDE.md
 
+# Runtime state
+control/
+
 # Staging area for temp files
 staging/
+
+# Python lock file
+uv.lock
 
 .serena/


.gitignore
```diff
@@ -1,7 +1,8 @@
 # Binaries
 /wtmcp
 /wtmcpctl
-plugins/*/handler
+plugins/google-*/handler
+plugins/gitlab/handler
```
sergio-correia, line 5: I wonder if something like `plugins/*/handler` would work better here, instead of listing plugins individually?

sergio-correia: And I just realized we do have `plugins/gitlab/handler` in the version we have in the master branch. Try to rebase, I think this PR is not needed.

thedeveloper-ctl:
Thanks @sergio-correia for the review! You're right — `main` already has the improved `plugins/*/handler` wildcard pattern (plus `uv.lock` and `control/`), which supersedes all changes in this PR. Confirmed by attempting a rebase on latest `main` which results in conflicts with no new content to add. Closing this PR as it's no longer needed. 👍

PR #16, opened 2026-03-31
fix: require word boundaries for wiki inline markup

decko:
## Summary
- Wiki-to-ADF inline formatting patterns (`-strike-`, `_italic_`, `*bold*`, etc.) matched mid-word delimiters, causing false positives on plain text like ISO dates (`2026-03-31` → ~~03~~), identifiers (`test_configuration_docs` → *configuration*), and hyphenated words (`foo-bar-baz` → ~~bar~~)
- Added `(?<!\w)` / `(?!\w)` word-boundary guards to all six inline formatting regex patterns so they only fire when delimiters sit at actual word boundaries (whitespace, punctuation, or start/end of text)
- Legitimate wiki markup (e.g., `some -struck- here`, `(_italic_)`) continues to work correctly

## Test plan
- [x] All 435 existing Jira plugin tests pass (0 regressions)
- [x] 10 new tests added covering the reported false-positive patterns and confirming proper word-boundary matching still works
- [x] Pre-commit hooks pass (ruff, ty, pytest, trailing whitespace, etc.)

### DIFF
--- a/plugins/jira/helpers.py
+++ b/plugins/jira/helpers.py
@@ -51,12 +51,12 @@ _WIKI_INLINE_RE = re.compile(
     r"|(?P<bare_link>\[(?P<bare_url>https?://[^]]+)\])"
     r"|(?P<mention>\[~(?P<mention_id>[^]]+)\])"
     r"|(?P<image>!(?P<image_url>\S+?\.\S+?)!)"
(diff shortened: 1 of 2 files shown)


sergio-correia (approved):
LGTM, thanks!

PR #17, opened 2026-04-06
pre-commit: fix hard-coded paths

rjeffman:
pre-commit hooks used hard-coded paths based on '.venv', which might not be the ones used on different hosts. This patch allows for different virtual environment paths.


sergio-correia (approved):
LGTM, thanks!

PR #18, opened 2026-04-06
Add 'pip' requirements for development

rjeffman:
Add a 'requirements-dev.txt' to make it easier to create a development environment using Python's 'pip' and virtual environments.


requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+pytest
+ty
+ruff
```
sergio-correia: Lets pin it to the version we have ein the pre-commit config, `ruff==0.11.13`; also, please add `pre-commit` to this list of dev dependencies.

rjeffman: I pinned all versions to what we have today.

sergio-correia (approved):
LGTM

PR #19, opened 2026-04-07
Fix testing-farm information gathering and status report

bmeneg:
This PR makes fixes a few issues with relation to how testing-farm plugin gathered the artifacts and also the IP address for SSH during reservations.

* We now properly grab the artifacts URL from the `run` object in the API response instead from the root object.
* Instead of relying on parsing console log (which can sometimes be empty), we now pull the SSH IP directly from `provision/guests.yaml`. If that fails, it still falls back to checking the console log.
* If a reservation is already in the `complete` state, the plugin now explicitly tells you that the host has been returned and is no longer accessible, instead of giving a false positive that you can still SSH into it.

### DIFF
--- a/plugins/testing-farm/handler.py
+++ b/plugins/testing-farm/handler.py
@@ -253,8 +253,8 @@ def testing_farm_get_request(params):
     envs = body.get("environments_requested", [])
     env0 = envs[0] if envs else {}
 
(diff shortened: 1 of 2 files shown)


sergio-correia (approved):
LGTM, thanks!

PR #20, opened 2026-04-07
fix: preserve text formatting in google-docs plugin

rjeffman:
The Google Docs API resets run-level text properties (bold, italic, underline) when UpdateParagraphStyle is applied. Fixed by removing NORMAL_TEXT paragraph style for normal text (unnecessary, inherits document defaults) and deferring heading text styles until after paragraph style application so formatting overrides style defaults.


sergio-correia (approved):
LGTM, thanks

PR #23, opened 2026-04-16
Add support for pre-formatted text in Google Docs plugin

rjeffman:
This PR adds support for pre-formatted text to Google Docs plugin.

Both inline and code blocks are supported.

For code blocks, the basic mechanism to support language highlighting in the future was implemented.

### DIFF
--- a/plugins/google-docs/tools.go
+++ b/plugins/google-docs/tools.go
@@ -122,17 +122,43 @@ func extractMarkdown(doc *docs.Document) string {
 		return ""
 	}
 
-	for _, elem := range doc.Body.Content {
-		extractElementMarkdown(&md, elem)
+	// Detect code annotations before processing
+	annotations := detectCode(doc)
+
+	for i, elem := range doc.Body.Content {
+		extractElementMarkdown(&md, elem, annotations, i)
 	}
 
 	return md.String()
 }
 
 // extractElementMarkdown recursively converts document elements to Markdown.
-func extractElementMarkdown(sb *strings.Builder, elem *docs.StructuralElement) {
+func extractElementMarkdown(sb *strings.Builder, elem *docs.StructuralElement, annotations *CodeAnnotations, paraIndex int) {
 	if elem.Paragraph != nil {
 		para := elem.Paragraph
+
+		// Check if this is a code block
+		if annotations.CodeBlocks[paraIndex] {
+			// Extract plain text from code block (strip all formatting)
+			var codeText strings.Builder
+			for _, pe := range para.Elements {
+				if pe.TextRun != nil {
+					codeText.WriteString(pe.TextRun.Content)
+				}
+			}
+
+			text := codeText.String()
+			if text != "" {
+				sb.WriteString("```\n")
+				sb.WriteString(text)
+				// Ensure code block ends with newline before closing fence
+				if !strings.HasSuffix(text, "\n") {
+					sb.WriteString("\n")
+				}
+				sb.WriteString("```\n\n")
+			}
+			return
+		}
 
 		// Determine if this is a heading
 		var prefix, suffix string
@@ -159,21 +185,44 @@ func extractElementMarkdown(sb *strings.Builder, elem *docs.StructuralElement) {
 
 		// Extract text with formatting
 		var paraText strings.Builder
-		for _, pe := range para.Elements {
+		for runIndex, pe := range para.Elements {
 			if pe.TextRun != nil {
 				text := pe.TextRun.Content
 				style := pe.TextRun.TextStyle
 
-				// Apply text formatting
+				// Check if this run is inline code
+				isInline := annotations.InlineCode[makeInlineCodeKey(paraIndex, runIndex)]
+				if isInline {
+					// Inline code: wrap in single backticks
+					text = "`" + strings.TrimSpace(text) + "`"
+				}
 				if style != nil {
-					if style.Bold {
-						text = "**" + strings.TrimSpace(text) + "**"
-					}
-					if style.Italic {
-						text = "*" + strings.TrimSpace(text) + "*"
-					}
-					if style.Underline {
-						text = "__" + strings.TrimSpace(text) + "__"
+					// Apply formatting (works for both inline code and regular text)
+					if !isInline {
+						// Only trim for non-inline-code (inline code already trimmed above)
+						if style.Bold {
+							text = "**" + strings.TrimSpace(text) + "**"
+						}
+						if style.Italic {
+							text = "*" + strings.TrimSpace(text) + "*"
+						}
+						if style.Underline {
+							text = "__" + strings.TrimSpace(text) + "__"
+						}
+					} else {
+						// For inline code, apply formatting INSIDE the backticks
+						// Strip backticks, apply formatting, re-wrap
+						inner := text[1 : len(text)-1] // remove surrounding backticks
+						if style.Bold {
+							inner = "**" + inner + "**"
+						}
+						if style.Italic {
+							inner = "*" + inner + "*"
+						}
+						if style.Underline && (style.Link == nil || style.Link.Url == "") {
+							inner = "__" + inner + "__"
+						}
+						text = "`" + inner + "`"
 					}
 					if style.Link != nil && style.Link.Url != "" {
 						text = "[" + strings.TrimSpace(text) + "](" + style.Link.Url + ")"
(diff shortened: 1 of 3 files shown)


plugins/google-docs/tools.go
```diff
@@ -946,6 +1189,70 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64) []*
...
+					},
+					TextStyle: &docs.TextStyle{
+						WeightedFontFamily: &docs.WeightedFontFamily{
+							FontFamily: "Courier New",
+						},
+						Bold:      false,
+						Italic:    false,
+						Underline: false,
+					},
+					Fields: "weightedFontFamily,bold,italic,underline",
```
decko: 🔴 **Critical: Strikethrough not reset in code block `UpdateTextStyle`**

*(Found independently by both Go Specialist and Skeptical QE agents)*

The `Fields` mask includes `bold,italic,underline` but omits `strikethrough`. Everywhere else in the file (e.g. lines 1557, 1582), strikethrough is included alongside the other formatting flags. If a code block is inserted after a strikethrough segment, the strikethrough formatting will "stick" to the code block text because the API is not told to reset it.

```suggestion
					Fields: "weightedFontFamily,bold,italic,underline,strikethrough",
```

Also add `Strikethrough: false` to the `TextStyle` struct above (around line 1227) for consistency.

plugins/google-docs/tools.go
```diff
@@ -572,6 +621,130 @@ type markdownSegment struct {
...
+				isMonospaceFont(pe.TextRun.TextStyle.WeightedFontFamily.FontFamily) {
+				monoRuns++
+				monoIndices = append(monoIndices, j)
+			}
+		}
+
+		if totalRuns == 0 {
+			continue
+		}
+
```
decko, line 726: 🔴 **Critical: Consecutive monospace paragraphs produce separate code fences**

*(Found independently by both Go Specialist and Skeptical QE agents)*

Google Docs stores each line of a code block as a **separate paragraph** element. `detectCode` marks each paragraph individually, and `extractElementMarkdown` wraps each one independently in triple-backtick fences.

For a document with three consecutive Courier New paragraphs (`line1`, `line2`, `line3`), the output will be:
```
` ` `
line1
` ` `

` ` `
line2
` ` `

` ` `
line3
` ` `
```

instead of a single merged code block.

**Suggestion:** Add a post-processing pass that merges consecutive `CodeBlocks` entries, or handle it in the rendering phase by tracking whether the previous paragraph was also a code block.

plugins/google-docs/tools.go
```diff
@@ -572,6 +621,130 @@ type markdownSegment struct {
 	dateValue         string // specific date in YYYY-MM-DD format, empty means @today
 	isPersonField     bool   // true if this should be a person field (@(name or email))
 	personIdentifier  string // name or email for person field
+	isInlineCode      bool   // true if text should be wrapped in single backticks
+	isCodeBlock       bool   // true if text is part of a code block (triple backticks)
+	codeLanguage      string // language identifier for code block (e.g., ```language syntax)
+}
+
+// CodeAnnotations holds code detection results for a document.
+type CodeAnnotations struct {
```
decko, line 630: 🟠 **High: `CodeAnnotations` should not be exported**

`CodeAnnotations` and its fields (`InlineCode`, `CodeBlocks`, `Languages`) are exported but only used internally. Both `detectCode` (which returns it) and `extractElementMarkdown` (which consumes it) are unexported. Exporting this type expands the public API surface unnecessarily and creates a maintenance commitment.

**Suggestion:** Rename to `codeAnnotations` with fields `inlineCode`, `codeBlocks`, `languages`.

plugins/google-docs/tools.go
```diff
@@ -572,6 +621,130 @@ type markdownSegment struct {
...
+
+	// Maps paragraph index to code block status
+	CodeBlocks map[int]bool
+
+	// Maps paragraph index to language hint (future use)
+	Languages map[int]string
+}
+
+// makeInlineCodeKey creates a composite key for inline code annotation.
+func makeInlineCodeKey(paraIndex, runIndex int) string {
```
decko, line 642: 🟠 **High: String-keyed map is fragile; struct key is more idiomatic Go**

*(Found independently by both agents)*

Using `fmt.Sprintf("%d:%d", ...)` for map keys is allocation-heavy and fragile (reversed parameters or different separators would silently fail). A struct key is more idiomatic:

```go
type textRunKey struct {
    paraIndex int
    runIndex  int
}

type codeAnnotations struct {
    inlineCode map[textRunKey]bool
    codeBlocks map[int]bool
    languages  map[int]string
}
```

This eliminates the `makeInlineCodeKey` function entirely and makes lookups self-documenting.

plugins/google-docs/tools.go
```diff
@@ -606,9 +779,48 @@ func parseMarkdown(markdown string) []markdownSegment {
...
+	// Code block state tracking
+	inCodeBlock := false
+	var codeBlockLines []string
+	var codeLanguage string
+
 	for _, line := range lines {
 		trimmedLine := strings.TrimSpace(line)
 
+		// Check for code fence boundaries (``` with optional language)
+		if strings.HasPrefix(trimmedLine, "```") {
```
decko, line 782: 🟠 **High: Indented triple-backtick lines inside code blocks prematurely close the block**

The fence check uses `trimmedLine` (whitespace-stripped), so a line like `"    ` ` `"` inside a code block will be interpreted as a closing fence. In standard Markdown (CommonMark), code fence closing must not be indented more than 3 spaces.

**Triggering scenario:** A code block containing Python with docstrings that include example code fences. The indented ` ``` ` would close the outer block prematurely.

**Suggestion:** Use `line` (not `trimmedLine`) for the fence check when inside a code block, or limit to at most 3 leading spaces per CommonMark spec.

plugins/google-docs/tools.go
```diff
@@ -704,6 +926,24 @@ func parseSimpleFormattingWithDepth(text string, depth int) []markdownSegment {
...
 
 	for pos < len(text) {
+		// Check for `inline code` (single backtick, not triple)
+		if text[pos] == '`' && !strings.HasPrefix(text[pos:], "```") {
+			endPos := strings.Index(text[pos+1:], "`")
+			if endPos != -1 {
+				endPos += pos + 1
+				// Parse formatting INSIDE backticks
+				codeText := text[pos+1 : endPos]
+				innerSegs := parseSimpleFormattingWithDepth(codeText, depth+1)
```
decko, line 927: 🟠 **High: Formatting is parsed inside inline code, violating Markdown spec**

The parser recursively calls `parseSimpleFormattingWithDepth` on content inside backticks, treating `**`, `*`, `__`, and `~~` as formatting markers. Standard Markdown treats backtick content as **verbatim** — no formatting is applied.

Input `` `**not bold**` `` produces a bold inline code segment instead of literal text. There is no way to include literal asterisks or underscores inside inline code.

This appears to be **intentional** for Google Docs round-trip compatibility (monospace text can also be bold in Docs), but it should be explicitly documented as a known deviation from the Markdown spec.

plugins/google-docs/tools.go
```diff
@@ -572,6 +621,130 @@ type markdownSegment struct {
...
+	"Menlo",
+	"Source Code Pro",
+	"SF Mono",
+	"Inconsolata",
+	"Roboto Mono",
+}
+
+// isMonospaceFont checks if the given font family is a monospace font.
+// Comparison is case-insensitive.
+func isMonospaceFont(fontFamily string) bool {
```
decko, line 659: 🟡 **Medium: `isMonospaceFont` re-lowercases constant data on every call**

The `monospaceFonts` slice contains compile-time constants, yet `strings.ToLower(mono)` is called on each element every invocation. For every text run in the document, this does 9 allocations.

**Suggestion:** Use a pre-computed `map[string]bool`:
```go
var monospaceFontSet = func() map[string]bool {
    fonts := []string{"courier new", "courier", "consolas", "monaco", "menlo",
        "source code pro", "sf mono", "inconsolata", "roboto mono"}
    m := make(map[string]bool, len(fonts))
    for _, f := range fonts {
        m[f] = true
    }
    return m
}()

func isMonospaceFont(fontFamily string) bool {
    if fontFamily == "" {
        return false
    }
    return monospaceFontSet[strings.ToLower(fontFamily)]
}
```

plugins/google-docs/tools.go
```diff
@@ -572,6 +621,130 @@ type markdownSegment struct {
...
+		for j, pe := range para.Elements {
+			if pe.TextRun == nil {
+				continue
+			}
+			// Skip runs that are only whitespace (e.g., trailing newlines).
+			content := strings.TrimSpace(pe.TextRun.Content)
+			if content == "" {
+				continue
+			}
+			totalRuns++
```
decko, line 714: 🟡 **Medium: Non-text paragraph elements cause false positive code block detection**

`detectCode` only counts `TextRun` elements. If a paragraph contains one inline image (`InlineObjectElement`, not a `TextRun`) and one Courier New text run, then `totalRuns == 1` and `monoRuns == 1`, marking the entire paragraph as a code block.

**Triggering scenario:** A paragraph containing an inline image followed by a Courier New caption.

plugins/google-docs/tools.go
```diff
@@ -572,6 +621,130 @@ type markdownSegment struct {
...
+			if pe.TextRun.TextStyle != nil &&
+				pe.TextRun.TextStyle.WeightedFontFamily != nil &&
+				isMonospaceFont(pe.TextRun.TextStyle.WeightedFontFamily.FontFamily) {
```
decko, line 717: 🟡 **Medium: Document-level default monospace font causes false negatives**

`detectCode` requires `WeightedFontFamily` to be explicitly set on each text run. If a document's default paragraph style uses Courier New, text runs inherit this font but may have `WeightedFontFamily: nil` in the API response (the API only includes explicitly overridden properties). These runs will not be detected as monospace.

**Triggering scenario:** A Google Doc where the user has set Format → Paragraph styles → Normal text to Courier New.

plugins/google-docs/tools_test.go
```diff
@@ -1640,3 +1642,1930 @@ func TestSaveDocumentFile(t *testing.T) {
...
+				foundCourierNew = true
+			}
+		}
+		if !foundCourierNew {
+			t.Errorf("expected Courier New font for inline code in heading")
+		}
+	})
+}
+
+func TestRoundTrip_CodeBlock(t *testing.T) {
```
decko, line 3410: 🟡 **Medium: Round-trip test uses unrealistic mock document**

This test puts all code content into a **single paragraph** with multiple text runs. In reality, when you write a code block via the Google Docs API (as `InsertText` with `\n`-delimited text), Google Docs splits it into **multiple paragraphs** (one per line). A subsequent read will hit the consecutive-paragraphs issue (each becoming its own ` ``` ` block).

**Suggestion:** Add a test variant that uses one `StructuralElement` per code line to simulate real Google Docs behavior, and verify they merge into a single code fence.

decko (commented):
## Combined Review: Go Specialist + Skeptical Quality Engineer

Two specialized review agents analyzed this PR independently. Here is a consolidated review with findings mapped to specific lines.

### Summary

The PR is well-structured with a clear separation between detection (`detectCode`), parsing (`parseMarkdown`), and rendering (`convertMarkdownToRequests`). Test coverage is strong at 1929 new lines. However, there are some issues that should be addressed before merging.

### Severity Legend
- 🔴 **Critical** — Will produce incorrect output in production
- 🟠 **High** — Significant correctness or design issue
- 🟡 **Medium** — Robustness concern
- 🔵 **Low/Suggestion** — Code quality improvement

### PR Title
Typo in PR title: "supoprt" → "support"

### What Both Agents Agreed Was Done Well
- Clean separation of concerns between detection, parsing, and rendering
- Defensive nil checks throughout
- Font reset on non-code segments prevents style bleeding
- Heading exclusion from code detection
- Unclosed code fence graceful handling
- Depth limit on recursive parsing prevents stack overflow

plugins/google-docs/tools.go
rjeffman: This is not an issue. I don't really expect people to "convert" to Markdown as much as "convert to GDocs", also this is unsolvable due to the way Google Docs structures text (actually "unstructured", only formatted).

plugins/google-docs/tools.go
rjeffman: This is by design.

plugins/google-docs/tools.go
rjeffman: There's no support for images.

plugins/google-docs/tools.go
rjeffman: This will be used by a "under implementation" tool to add syntax highlighting to code blocks.

plugins/google-docs/tools.go
rjeffman: This is a non-issue.

plugins/google-docs/tools.go
rjeffman: We're using a loose interpretation of the Markdown rules. Let's wait for it to become a real issue.

plugins/google-docs/tools.go
rjeffman: Fixed.

plugins/google-docs/tools.go
rjeffman: Let's wait for it to became an issue.

plugins/google-docs/tools.go
rjeffman: Fixed

plugins/google-docs/tools_test.go
rjeffman: The text reproduces an issue that was happening.

rjeffman:
@decko thanks for the review. Items were either addressed or answered.

sergio-correia (approved):
LGTM, thanks

PR #7, opened 2026-03-19
feat(jira): add set_parent and clear_parent tools

thedeveloper-ctl:
## Summary

- Add `jira_set_parent` tool: set the parent of one or more issues across any hierarchy level (Outcome → Feature → Epic → Story/Task). Tries the standard REST API first; if screen restrictions prevent it, automatically falls back to the Jira Agile API (`POST /rest/agile/1.0/epic/{key}/issue`).
- Add `jira_clear_parent` tool: remove the parent from one or more issues using the same dual-strategy approach (REST API + Agile "none" epic fallback).
- Both tools follow the existing `dry_run=true` default pattern and comma-separated `issue_keys` input format.

## Motivation

Setting parent/epic links via `jira_set_custom_field` fails on many Jira Cloud instances because the Parent Link field (`customfield_10018`) is not on the issue edit screen. The Agile API endpoint bypasses these screen restrictions entirely, making it the reliable way to manage epic-child relationships.

## Changes

| File | Change |
|------|--------|
| `plugins/jira/tools_write.py` | Add `set_parent()` and `clear_parent()` functions + register in `TOOLS` dict |
| `plugins/jira/plugin.yaml` | Add tool definitions with params and descriptions |
| `plugins/jira/tests/test_tools_write.py` | Add 11 unit tests (dry runs, REST success, Agile fallback, partial failures, validation) |

## Test plan

- [x] All 60 unit tests pass (`pytest tests/test_tools_write.py` — 60 passed)
- [ ] Manual test: `jira_set_parent` with dry_run=true returns preview
- [ ] Manual test: `jira_set_parent` with dry_run=false links child to epic
- [ ] Manual test: `jira_clear_parent` removes parent link

PR #25, opened 2026-04-21
Add support for tables in Google Docs

rjeffman:
Add support for adding tables to Google Docs. Tables are created and
then the cell content is adde one by one. Each step requires a batch
execution of the Google Docs API.
    
The format support is the same as standard Markdown tables, no cell
merging is supported. Table headers are written in bold by default.
    
An extension to the Markdown table format implement is to render tables
without the table header by starting the table definition with the
header separator as in:

``` 
|----|----|
| A1 | A2 |
| B1 | B2 |
```

The resulting table is rendered without a table header.

### DIFF
--- a/plugins/google-docs/tools.go
+++ b/plugins/google-docs/tools.go
@@ -26,6 +26,9 @@ var (
 	reDateChip      = regexp.MustCompile(`@date\((\d{4}-\d{2}-\d{2})\)`)
 	rePersonChip    = regexp.MustCompile(`@\(([^)]+)\)`)
 	reLinkInline    = regexp.MustCompile(`\[([^\]]+)\]\(([^\)]+)\)`)
+	// Table detection patterns
+	reTableRow       = regexp.MustCompile(`^\s*\|(.+\|)+\s*$`)
+	reTableSeparator = regexp.MustCompile(`^\s*\|[\s\-:]*\-[\s\-:]*(\|[\s\-:]*\-[\s\-:]*)*\|\s*$`)
 )
 
 // extractDocumentID extracts a Google Docs document ID from a URL.
@@ -624,6 +627,26 @@ type markdownSegment struct {
 	isInlineCode      bool   // true if text should be wrapped in single backticks
 	isCodeBlock       bool   // true if text is part of a code block (triple backticks)
 	codeLanguage      string // language identifier for code block (e.g., ```language syntax)
+	// Table support
+	isTable bool
+	table   *tableSegment // Non-nil when isTable=true
+}
+
+// tableCell represents a single cell in a markdown table.
+type tableCell struct {
+	segments []markdownSegment // Cell content as formatted segments
+}
+
+// tableRow represents a single row in a markdown table.
+type tableRow struct {
+	cells []tableCell
+}
+
+// tableSegment represents a parsed markdown table.
+type tableSegment struct {
+	rows       []tableRow
+	numColumns int
+	isTable    bool // Always true, used for type discrimination
 }
 
 // CodeAnnotations holds code detection results for a document.
@@ -775,11 +798,31 @@ func parseMarkdown(markdown string) []markdownSegment {
 	var codeBlockLines []string
 	var codeLanguage string
 
+	// Table buffering state
+	var tableBuffer []string
+	inTable := false
+
 	for _, line := range lines {
 		trimmedLine := strings.TrimSpace(line)
 
 		// Check for code fence boundaries (``` with optional language)
 		if strings.HasPrefix(trimmedLine, "```") {
+			// Close any open table before code block
+			if inTable {
+				inTable = false
+				if table := parseMarkdownTable(tableBuffer); table != nil {
+					segments = append(segments, markdownSegment{
+						isTable: true,
+						table:   table,
+					})
+				} else {
+					for _, bufferedLine := range tableBuffer {
+						segments = append(segments, parseSimpleFormatting(bufferedLine+"\n")...)
+					}
+				}
+				tableBuffer = nil
+			}
+
 			if !inCodeBlock {
 				// Opening fence: start accumulating code block lines
 				inCodeBlock = true
@@ -812,10 +855,42 @@ func parseMarkdown(markdown string) []markdownSegment {
 			continue
 		}
 
+		// Check if this line looks like a table row
+		isTableRow := reTableRow.MatchString(line)
+
+		if isTableRow {
+			// Start or continue buffering table lines
+			if !inTable {
+				inTable = true
+				tableBuffer = []string{}
+			}
+			tableBuffer = append(tableBuffer, line)
+			continue
+		}
+
+		// Not a table row - process any buffered table
+		if inTable {
+			inTable = false
+			// Try to parse buffered lines as a table
+			if table := parseMarkdownTable(tableBuffer); table != nil {
+				// Valid table
+				segments = append(segments, markdownSegment{
+					isTable: true,
+					table:   table,
+				})
+			} else {
+				// Invalid table - treat as plain text paragraphs
+				for _, bufferedLine := range tableBuffer {
+					segments = append(segments, parseSimpleFormatting(bufferedLine+"\n")...)
+				}
+			}
+			tableBuffer = nil
+		}
+
+		// Process non-table line normally
 		if trimmedLine == "" {
 			if lastWasHeading {
-				// Skip blank lines immediately after headings —
-				// headings create their own paragraph break via \n
+				// Skip blank lines immediately after headings
 				continue
 			}
 			// Preserve blank lines between normal text as empty paragraphs
@@ -827,12 +902,12 @@ func parseMarkdown(markdown string) []markdownSegment {
 		headingLevel := 0
 		if strings.HasPrefix(trimmedLine, "#") {
 		loop:
-			for i, ch := range trimmedLine {
+			for j, ch := range trimmedLine {
 				switch ch {
 				case '#':
 					headingLevel++
 				case ' ':
-					trimmedLine = trimmedLine[i+1:]
+					trimmedLine = trimmedLine[j+1:]
 					break loop
 				default:
 					headingLevel = 0
@@ -843,14 +918,12 @@ func parseMarkdown(markdown string) []markdownSegment {
 
 		if headingLevel > 0 && headingLevel <= 6 {
 			lastWasHeading = true
-			// Heading line - parse inline formatting within heading text
-			// Note: Headings don't include trailing newline - they're standalone paragraphs
 			inlineSegs := parseSimpleFormatting(trimmedLine)
 			lineID := nextHeadingLineID
 			nextHeadingLineID++
-			for i := range inlineSegs {
-				inlineSegs[i].heading = headingLevel
-				inlineSegs[i].headingLineID = lineID
+			for j := range inlineSegs {
+				inlineSegs[j].heading = headingLevel
+				inlineSegs[j].headingLineID = lineID
 			}
 			segments = append(segments, inlineSegs...)
 			continue
@@ -858,27 +931,27 @@ func parseMarkdown(markdown string) []markdownSegment {
 
 		lastWasHeading = false
 
-		// Check for ordered list (e.g., "1. Item", "2. Item") with optional leading indent
+		// Check for ordered list
 		if orderedListMatch := reOrderedList.FindStringSubmatch(line); orderedListMatch != nil {
 			depth := indentDepth(orderedListMatch[1])
 			listItemText := orderedListMatch[2]
 			inlineSegs := parseSimpleFormatting(listItemText + "\n")
-			for i := range inlineSegs {
-				inlineSegs[i].orderedListItem = true
-				inlineSegs[i].listDepth = depth
+			for j := range inlineSegs {
+				inlineSegs[j].orderedListItem = true
+				inlineSegs[j].listDepth = depth
 			}
 			segments = append(segments, inlineSegs...)
 			continue
 		}
 
-		// Check for unordered list (e.g., "- Item", "* Item", "+ Item") with optional leading indent
+		// Check for unordered list
 		if unorderedListMatch := reUnorderedList.FindStringSubmatch(line); unorderedListMatch != nil {
 			depth := indentDepth(unorderedListMatch[1])
 			listItemText := unorderedListMatch[3]
 			inlineSegs := parseSimpleFormatting(listItemText + "\n")
-			for i := range inlineSegs {
-				inlineSegs[i].unorderedListItem = true
-				inlineSegs[i].listDepth = depth
+			for j := range inlineSegs {
+				inlineSegs[j].unorderedListItem = true
+				inlineSegs[j].listDepth = depth
 			}
 			segments = append(segments, inlineSegs...)
 			continue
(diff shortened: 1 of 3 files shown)


decko (commented):
## Comprehensive PR Review Summary

**Reviewed by 8 specialist agents:** Code Reviewer, Test Analyzer, Comment Analyzer, Silent Failure Hunter, Type Design Analyzer, Go Specialist, Google Docs Specialist, Google API Specialist

---

### Critical Issues (2)

1. **`toolWriteMarkdown` does not handle tables — creates empty tables silently** _(found by 7/8 agents)_
   - The multi-phase table logic (create → query → populate cells) only exists in `toolWriteText`. `toolWriteMarkdown` calls `convertMarkdownToRequests` directly, which emits `InsertTable` but never populates cell content. Users get empty tables with `"status": "success"`.
   - **Fix:** Extract shared multi-phase logic into a helper used by both tools.

2. **`break` in row/col loops silently drops cells on dimension mismatch** (tools.go:2091-2098)
   - When the API returns fewer rows/columns than requested, `break` skips remaining cells while returning `"status": "success"`. Should return an error.

### High Severity (8)

3. **O(R×C) API calls per table** — A 10×5 table makes ~102 API calls. Google Docs quota is ~300 writes/min. Fix: batch all cell writes into one `BatchUpdate` (cells are isolated — inserting in one doesn't shift others). Reduces to 3 API calls total.
4. **`elem` retains stale value when re-find loop finds nothing** (tools.go:2122) — Missing `elem = nil` reset before scan.
5. **Table index size estimation is wrong** (tools.go:1483) — Formula uses desired content sizes but table is created empty. Actual empty R×C table = `2 + 2×R×C` index positions.
6. **Smart chips degraded to plain text in table cells** (tools.go:1364) — Uses `InsertText` instead of `InsertDate`/`InsertPerson`. The API supports smart chips in cells. Contradicts context.md documentation.
7. **No nil guard on `doc.Body` after re-queries** — The ~6 re-queries during cell population never check `doc.Body != nil`.
8. **Non-table segment index tracking misses `InsertDate`/`InsertPerson`** (tools.go:2166) — Only counts `InsertText` for index advancement.
9. **No concurrent modification protection** — API supports `WriteControl.RequiredRevisionId` but code doesn't use it.
10. **`currentIndex` not updated if table not found in final re-query** (tools.go:2140-2150)

### Medium Severity (6)

11. Misplaced test subtests — inline code tests inside `TestTableEdgeCases` (line 4075), table tests inside `TestRoundTrip_InlineCode` (line 4327)
12. `tableSegment.isTable` field always true — dead weight (tools.go:649)
13. Empty cell space insertion unnecessary — cells already have default `\n` (tools.go:1346)
14. No table dimension validation — Google Docs supports ~20 columns max
15. Stale `tableRow` reference after re-query (tools.go:2133)
16. `reTableRow` nested quantifier `(.+\|)+` unnecessarily complex

### Test Coverage Gaps

| Priority | Gap |
|----------|-----|
| 1 | `populateTableCell` completely untested (100-line function, 5 branches) |
| 2 | Headerless tables (separator-first) completely untested |
| 3 | `toolWriteText` multi-phase orchestration untested |
| 4 | Header bold formatting never asserted |

### Strengths

- Solid parsing pipeline with proper fallback-to-plain-text for invalid tables
- Good error wrapping with descriptive context including table/cell coordinates
- Two-phase API approach is architecturally correct
- Rich text in cells well-tested at parse level
- Goroutine-safe, good pre-allocation patterns

### Top Recommendations

1. **Batch all cell writes into one `BatchUpdate`** — fixes issues #3, #4, #7, #8, #10 simultaneously
2. **Add table support to `toolWriteMarkdown`** — extract shared helper
3. **Use `InsertDate`/`InsertPerson` in `populateTableCell`** — API supports it in cells
4. **Fix silent `break` to return errors** — dimension mismatches are faults
5. **Fix misplaced test subtests**

---
_Review generated by 8 specialized AI agents (Code Reviewer, Test Analyzer, Comment Analyzer, Silent Failure Hunter, Type Design Analyzer, Go Specialist, Google Docs Specialist, Google API Specialist)_

plugins/google-docs/tools.go
```diff
@@ -1718,6 +2040,148 @@ func toolWriteText(params, _ json.RawMessage) (any, error) {
...
+						}
+					}
+
+					if elem == nil || elem.Table == nil {
+						return nil, fmt.Errorf("table %d not found after creation", tableIdx)
+					}
+
+					// Populate each cell in a separate batch
+					for rowIdx, row := range seg.table.rows {
+						if rowIdx >= len(elem.Table.TableRows) {
```
decko: **Critical: Silent cell dropping.** When the Google Docs API returns a table with fewer rows than requested, this `break` silently skips all remaining rows. The function returns `"status": "success"` — the user has no idea cells were lost.

```suggestion
						if rowIdx >= len(elem.Table.TableRows) {
							return nil, fmt.Errorf("table %d: expected %d rows but API returned %d", tableIdx, len(seg.table.rows), len(elem.Table.TableRows))
						}
```

Same issue applies to the column `break` at line 2098.

plugins/google-docs/tools.go
```diff
@@ -1718,6 +2040,148 @@ func toolWriteText(params, _ json.RawMessage) (any, error) {
...
+							elem = e
+							break
+						}
+					}
+
+					if elem == nil || elem.Table == nil {
+						return nil, fmt.Errorf("table %d not found after creation", tableIdx)
+					}
+
+					// Populate each cell in a separate batch
```
decko: **High: O(R×C) API calls per table.** Each cell gets its own `BatchUpdate` + `Documents.Get` round-trip. A 10×5 table = 102 API calls (quota is ~300/min).

Since table cells are isolated structural elements, inserting text in one cell does NOT shift indices in other cells. After the initial `Documents.Get` (line 2071), you can collect ALL cell population requests and execute them in a **single** `BatchUpdate`. This reduces any table to 3 API calls total and makes cell population atomic.

This also eliminates the stale-`elem` bug at line 2122 and the missing nil guards on `doc.Body` after re-queries.

plugins/google-docs/tools.go
```diff
@@ -1718,6 +2040,148 @@ func toolWriteText(params, _ json.RawMessage) (any, error) {
...
+									return nil, fmt.Errorf("populate table %d cell[%d,%d]: %w", tableIdx, rowIdx, colIdx, err)
+								}
+
+								// Re-query document after each cell
+								doc, err = docsSvc.Documents.Get(docID).Do()
+								if err != nil {
+									return nil, fmt.Errorf("get document after table %d cell[%d,%d]: %w", tableIdx, rowIdx, colIdx, err)
+								}
+
+								// Re-find the table at currentIndex
```
decko: **High: `elem` retains stale value.** This loop scans for the table but doesn't reset `elem = nil` before scanning. If no table is found at or after `currentIndex`, the loop exits without setting `elem`, and the nil guard at line 2129 never fires because `elem` still holds its previous (stale) value. Subsequent cell writes would use wrong indices.

```suggestion
								// Re-find the table at currentIndex
								elem = nil
								for _, e := range doc.Body.Content {
```

plugins/google-docs/tools.go
```diff
@@ -1176,6 +1474,30 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64) []*
 	var currentHeadingLineID int
 
 	for i, seg := range segments {
+		// Handle table segments
+		if seg.isTable && seg.table != nil {
+			tableRequests := convertTableToRequests(seg.table, currentIndex)
+			requests = append(requests, tableRequests...)
+
+			// Advance index past the table
+			// Table size: 1 char (table marker) + sum of all cell content + cell boundaries
```
decko: **High: Table size estimation is wrong.** This formula estimates based on *desired cell content*, but the table is created *empty* by `InsertTable`. In the Google Docs model, an empty R×C table occupies `2 + 2×R×C` index positions (1 table start + R×C cells × 2 positions each + 1 trailing newline).

This formula is only used by `convertMarkdownToRequests`, which is called directly from `toolWriteMarkdown` — so any content after a table in `toolWriteMarkdown` will be inserted at the wrong index.

Also: `toolWriteMarkdown` never populates cell content (the multi-phase logic only exists in `toolWriteText`), so tables via that path will be empty. The shared table logic should be extracted into a helper.

plugins/google-docs/tools.go
```diff
@@ -1141,6 +1324,121 @@ func mergeSegments(segments []markdownSegment) []markdownSegment {
...
+	}
+
+	// Insert cell content segment by segment
+	currentIndex := cellStartIndex
+	for _, seg := range cell.segments {
+		if seg.text == "" && !seg.isDateField && !seg.isPersonField {
+			continue
+		}
+
+		// Handle date chips
```
decko, line 1566: **High: Smart chips degraded to plain text.** `populateTableCell` uses `InsertText` for date chips (and person chips at line 1381), while the main `convertMarkdownToRequests` path uses `InsertDate`/`InsertPerson` API requests to create interactive smart chips.

The Google Docs API **does** support `InsertDate` and `InsertPerson` inside table cells — the only requirement is insertion inside a paragraph, which cells have by default.

`context.md` explicitly states "Date and person chips work in table cells" with examples, but the actual result is plain text (e.g., `2026-05-01` instead of an interactive date chip).

plugins/google-docs/tools.go
```diff
@@ -624,6 +627,26 @@ type markdownSegment struct {
...
+// tableRow represents a single row in a markdown table.
+type tableRow struct {
+	cells []tableCell
+}
+
+// tableSegment represents a parsed markdown table.
+type tableSegment struct {
+	rows       []tableRow
+	numColumns int
+	isTable    bool // Always true, used for type discrimination
```
decko, line 649: **Medium: Redundant field.** `isTable` is always `true` by construction — it's set to `true` in `parseMarkdownTable` and never checked for any decision. Type discrimination is already handled by `markdownSegment.isTable` + the nil check on `markdownSegment.table`. This field carries no information and may be flagged by golangci-lint's `unused` rule.

Consider removing it.

plugins/google-docs/tools.go
```diff
@@ -1718,6 +2040,148 @@ func toolWriteText(params, _ json.RawMessage) (any, error) {
...
+					if len(segRequests) > 0 {
+						batchUpdateReq := &docs.BatchUpdateDocumentRequest{Requests: segRequests}
+						_, err := docsSvc.Documents.BatchUpdate(docID, batchUpdateReq).Do()
+						if err != nil {
+							return nil, fmt.Errorf("insert content: %w", err)
+						}
+
+						// Update currentIndex
+						for _, req := range segRequests {
+							if req.InsertText != nil {
```
decko: **High: Missing index tracking for smart chips.** This only counts `InsertText` requests for `currentIndex` advancement. `convertMarkdownToRequests` also generates `InsertDate` and `InsertPerson` requests (each advances the index by 1). If a non-table segment between two tables contains `@today` or `@(email)`, the index will be wrong for all subsequent content.

```suggestion
							if req.InsertText != nil {
								currentIndex += int64(utf8.RuneCountInString(req.InsertText.Text))
							} else if req.InsertDate != nil {
								currentIndex += 1
							} else if req.InsertPerson != nil {
								currentIndex += 1
							}
```

plugins/google-docs/tools.go
```diff
@@ -1141,6 +1324,121 @@ func mergeSegments(segments []markdownSegment) []markdownSegment {
...
+			},
+		},
+	}
+}
+
+// populateTableCell creates requests to populate a single table cell with content.
+func populateTableCell(cell *tableCell, cellStartIndex int64) []*docs.Request {
+	var requests []*docs.Request
+
+	// Handle empty cells - insert a single space
```
decko: **Medium: Unnecessary space in empty cells.** Google Docs table cells already contain a default empty paragraph with a `\n` when created via `InsertTable`. Inserting `" "` produces cells with `" \n"` (visible whitespace) rather than truly empty cells. Consider skipping empty cells entirely.

plugins/google-docs/tools.go
```diff
@@ -1718,6 +2040,148 @@ func toolWriteText(params, _ json.RawMessage) (any, error) {
...
+								}
+								// Update tableRow reference
+								if rowIdx < len(elem.Table.TableRows) {
+									tableRow = elem.Table.TableRows[rowIdx]
+								}
+							}
+						}
+					}
+
+					// Update currentIndex to after this table for next segment
```
decko: **High: Silent index corruption.** If the table-finding loop (lines 2145-2150) doesn't find a table at `currentIndex`, `currentIndex` is never updated. Subsequent segments will be inserted at wrong positions. No error is returned — the function continues and reports success.

```suggestion
					// Update currentIndex to after this table for next segment
					doc, err = docsSvc.Documents.Get(docID).Do()
					if err != nil {
						return nil, fmt.Errorf("get document after completing table %d: %w", tableIdx, err)
					}
					tableFound := false
					for _, e := range doc.Body.Content {
						if e.Table != nil && e.StartIndex >= currentIndex {
							currentIndex = e.EndIndex
							tableFound = true
							break
						}
					}
					if !tableFound {
						return nil, fmt.Errorf("table %d not found when computing end index", tableIdx)
					}
```

plugins/google-docs/tools_test.go
```diff
@@ -3568,4 +4324,48 @@ func TestRoundTrip_InlineCode(t *testing.T) {
 
 	// Also verify original markdown parses correctly
 	_ = parseMarkdown(originalMarkdown)
+	t.Run("malformed table - inconsistent columns", func(t *testing.T) {
```
decko: **Medium: Misplaced subtests.** These two table-related subtests (`"malformed table - inconsistent columns"` and `"table with smart chips"`) are nested inside `TestRoundTrip_InlineCode`. They should be in `TestTableEdgeCases` or a dedicated table test function.

Similarly, at line 4075, `"bold inline code preserves bold formatting"` and related inline code subtests ended up inside `TestTableEdgeCases` due to brace misalignment — they should be restored to `TestConvertMarkdownToRequests_InlineCode`.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: Fixed.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: This is a hard one. Without the O(RxC) calls the text is added only to the first cell of the table. There's too many variations in the computation of the text indexes that, as of today, the plugin can't do so it requests the document again.

Unfortunately, that's what's working now.

plugins/google-docs/tools_test.go
rjeffman: fixed.

plugins/google-docs/tools.go
rjeffman: Will keep it for now.

sergio-correia (approved):
LGTM, thanks!

PR #30, opened 2026-05-01
feat: add keyring credential management system

rjeffman:
Implements OS-native keyring integration for secure credential storage with full backward compatibility.                                                                                     
                                                                                                          
  - **OS-native keyring integration**: Linux Secret Service, macOS Keychain                                                                                                                    
  - **OAuth token encryption**: AES-256-GCM with access tokens in keyring                                                                                                                      
  - **Encrypted backup/restore**: Argon2id key derivation for offline backups                                                                                                                  
  - **Migration support**: Atomic migration from env.d to keyring with rollback                                                                                                                
  - **CLI tools**: 12 new `wtmcpctl credentials` commands                                                                                                                                      
  - **Graceful degradation**: Falls back to env.d/envvar when keyring unavailable


internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	log.Printf("[credentials] migrated plaintext token to encrypted: %s/%s", group, pluginName)
+	return nil
+}
+
+// getOrCreateEncryptionKey retrieves or generates the AES-256
+// encryption key for a credential group. Keys are stored
+// base64-encoded in the keyring.
+func (te *TokenEncryption) getOrCreateEncryptionKey(group string) ([]byte, error) {
```
sergio-correia, line 250: `getOrCreateEncryptionKey` does a non-atomic check-then-create: reads the keyring, and if absent, generates a new key and stores it. `TokenEncryption` has no mutex.

Two OAuth2 plugins sharing a credential group (e.g., `google-calendar` and `google-drive`) can call `SaveToken` concurrently during token refresh. Both see `ErrCredentialNotFound`, both generate different keys, the second `Set` overwrites the first. The token encrypted with the lost key becomes permanently unrecoverable.

Please add a `sync.Mutex` to `TokenEncryption` and double-check after acquiring the lock:

```go
te.mu.Lock()
defer te.mu.Unlock()
// Re-check under lock
value, err = te.keyringStore.Get(group, "__encryption_key")
if err == nil {
    return base64.StdEncoding.DecodeString(value)
}
// Now safe to generate and store
```

Also: the decoded key length is not validated here, but `getEncryptionKey` (line 242) does check `len(key) != 32`. A corrupted 16-byte keyring entry would silently encrypt with AES-128, then `LoadToken` rejects it. Please add the same length check after the base64 decode on line 207.

internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	// Marshal the payload to JSON.
+	plaintext, err := json.Marshal(payload) //nolint:gosec // tokens are immediately encrypted with AES-256-GCM
+	if err != nil {
+		return nil, fmt.Errorf("marshal payload: %w", err)
+	}
+
+	// Encrypt with GCM. The sealed output is ciphertext + tag.
+	sealed := gcm.Seal(nil, nonce, plaintext, nil)
```
sergio-correia: `encryptPayload` and `decryptPayload` pass `nil` as the AAD to `gcm.Seal`/`gcm.Open`. All plugins in a group share the same encryption key, so an attacker with file-write access can swap `token-calendar.json.enc` and `token-drive.json.enc` within the same group and GCM decryption succeeds for both.

Contrast with `wtmcpvault/password.go` which correctly uses `BuildAAD(header, credentialGroup)` to prevent exactly this.

Since this hasn't been deployed yet, we can just add AAD without a migration path:

```go
aad := []byte(group + "\x00" + pluginName)
sealed := gcm.Seal(nil, nonce, plaintext, aad)
```

Same change in `decryptPayload`.

internal/credentials/backup.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	// Decode the salt.
+	salt, err := base64.StdEncoding.DecodeString(backup.KDFParams.Salt)
+	if err != nil {
+		return fmt.Errorf("decode salt: %w", err)
+	}
+
+	// Derive the encryption key using params from the backup.
+	key := deriveKey(
```
sergio-correia, line 202: `RestoreBackup` reads KDF parameters (`Memory`, `Iterations`, `Parallelism`) directly from the backup file and passes them to `argon2.IDKey` without bounds checking. A crafted backup file with `Memory: 4294967295` (~4TB) causes the process to OOM or hang.

`CreateBackup` correctly uses hardcoded safe parameters (64MB/3/4), but the restore path trusts whatever is in the file.

Please validate before calling `deriveKey`:

```go
if backup.KDFParams.Memory > 4*1024*1024 || // 4GB in KB
    backup.KDFParams.Iterations > 100 ||
    backup.KDFParams.Parallelism == 0 {
    return fmt.Errorf("KDF parameters out of safe range")
}
```

internal/credentials/backup.go
```diff
@@ -0,0 +1,374 @@
...
+// CreateBackup exports all migrated credentials to an encrypted
+// backup file. The backup is encrypted using AES-256-GCM with a key
+// derived from the password using Argon2id.
+//
+// If outputPath is empty, a default path is used:
+// ~/.config/wtmcp/credentials/backup-<timestamp>.enc
+//
+// The password slice is NOT zeroed by CreateBackup; callers are
+// responsible for zeroing it after the call returns.
+func (s *Service) CreateBackup(password []byte, outputPath string) error {
```
sergio-correia, line 86: Neither `CreateBackup` nor `RestoreBackup` acquire `s.mu`. Every other `Service` method that touches `keyringStore`, `migrationState`, or `cache` holds the lock. `RestoreBackup` mutates all three, so a concurrent `Get` could observe a partially-restored state (keys in keyring but group not yet marked migrated).

Currently these are CLI-only (separate process from the server), so it's not triggerable in practice. But the `Service` doc says "safe for concurrent use" and the methods are exported on `*Service`, so a future caller could reasonably assume they're safe.

Shall we either wrap `RestoreBackup` in `s.mu.Lock()` and `CreateBackup` in `s.mu.RLock()`, or document them as CLI-only and not concurrency-safe?

internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,336 @@
...
+		keyringStore: keyringStore,
+		tokensDir:    tokensDir,
+	}
+}
+
+// SaveToken encrypts and saves an OAuth token. The refresh token and metadata
+// are AES-256-GCM encrypted on disk first; only after a successful disk write
+// is the access token stored in the keyring. This ordering ensures that a disk
+// write failure never leaves a mismatched keyring/disk state.
+func (te *TokenEncryption) SaveToken(group, pluginName string, token *oauth2.Token) error {
```
sergio-correia, line 64: If the disk write (line 98) succeeds but the keyring write (line 104) fails, `SaveToken` returns an error without rolling back the disk file. On the next `LoadToken`, the encrypted payload decrypts fine but the access token read from the keyring fails, returning a hard error.

The comment on line 105 says "the token will be refreshed on next load" but `LoadToken` returns `fmt.Errorf("get access token: %w", err)`, not a partial token that would trigger a refresh.

Would it make sense to have `LoadToken` return a token with an empty `AccessToken` when the keyring read fails but the encrypted disk data is valid? That way the OAuth refresh flow kicks in automatically.

test/e2e/credentials_test.sh
```diff
@@ -0,0 +1,453 @@
...
+        return
+    fi
+
+    setup_workdir
+    trap cleanup_workdir EXIT
+
+    local output
+
+    # Set a credential with --keyring flag.
+    if ! output=$("${WTMCPCTL}" credentials set jira JIRA_TOKEN token123 --keyring 2>&1); then
```
sergio-correia: `credSetCmd` is defined with `cobra.ExactArgs(2)` (group + key), and prompts for the value interactively. This passes 3 positional args:

```bash
"${WTMCPCTL}" credentials set jira JIRA_TOKEN token123 --keyring
```

Cobra rejects this with "accepts 2 arg(s), received 3", so the set/get E2E path is untested. Same issue on line 192.

Please pipe via stdin instead:

```suggestion
    if ! output=$(echo "token123" | "${WTMCPCTL}" credentials set jira JIRA_TOKEN --keyring 2>&1); then
```

sergio-correia (commented):
Found some issues to address before merging, mostly around the token encryption lifecycle.

internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// Generate a new 256-bit key.
+	key := make([]byte, 32)
+	if _, err := rand.Read(key); err != nil {
+		return nil, fmt.Errorf("generate encryption key: %w", err)
+	}
+
+	// Store base64-encoded in the keyring.
+	encoded := base64.StdEncoding.EncodeToString(key)
+	if err := te.keyringStore.Set(group, "__encryption_key", encoded); err != nil {
```
sergio-correia, line 278: `__encryption_key` entries are stored directly via `te.keyringStore.Set()`, bypassing `svc.Set()` and `migrationState.TrackKey()`. This means `CreateBackup` (which reads from `migrationState.GetGroupKeys`) never includes the encryption keys.

After restoring a backup on a new machine, the encryption keys are missing from the keyring. All encrypted token files on disk become permanently undecryptable.

The fix would be to call `TrackKey(group, "__encryption_key")` here after the keyring write, so the key is included in backups. Same applies to `__oauth_access_*` entries stored by `SaveToken`.

test/e2e/credentials_test.sh
rjeffman: fixed.

internal/credentials/token_encryption.go
rjeffman: fixed.

internal/credentials/token_encryption.go
rjeffman: fixed.

internal/credentials/backup.go
rjeffman: fixed

internal/credentials/backup.go
rjeffman: fixed.

internal/credentials/token_encryption.go
rjeffman: fixed.

internal/credentials/token_encryption.go
rjeffman: fixed.

internal/credentials/backup.go
```diff
@@ -0,0 +1,404 @@
...
+	)
+	defer zeroBackupKey(key)
+
+	// Decrypt the payload.
+	payload, err := decryptBackupPayload(key, &backup)
+	if err != nil {
+		return fmt.Errorf("decrypt backup: %w", err)
+	}
+
+	// Import credentials.
```
sergio-correia: The restore loop writes credentials to the keyring one by one, but if a write fails mid-loop, the already-written keys remain with no rollback. The migration state ends up partially updated too.

`runCredMigrate` in `cred_ops.go:457-468` already handles this correctly — it tracks written keys and rolls them back on failure. Would be good to apply the same pattern here: track `(group, key)` pairs as they're written and delete them if a later write fails.

Also worth noting: the rollback path in `runCredMigrate` itself silently ignores delete errors (`_ = credService.DeleteFromKeyring(group, sk)`). If the keyring is down for writes, it's probably down for deletes too — reporting rollback failures would help the user know manual cleanup is needed.

internal/credentials/backup.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	return nil
+}
+
+// deriveKey derives an encryption key from a password using Argon2id.
+func deriveKey(password, salt []byte, memory, iterations uint32, parallelism uint8) []byte {
+	return argon2.IDKey(password, salt, iterations, memory, parallelism, argon2KeyLen)
+}
+
+func zeroBackupKey(key []byte) {
```
sergio-correia, line 319: This zeroing function is missing `runtime.KeepAlive(b)`, which means the compiler is allowed to optimize away the loop as a dead store. The three other `zeroBytes` implementations in `internal/secrets/` (decryptor.go:228, pinentry.go:318, wtmcpvault/password.go:280) all include `runtime.KeepAlive` for exactly this reason. Same issue in `cred_ops.go:1134` (`zeroSlice`).

Having 5 copies of essentially the same function — 3 correct, 2 broken — is a good sign it should be consolidated into one shared helper. Maybe `internal/secrets/zero.go` or similar, used everywhere.

```suggestion
func zeroBackupKey(key []byte) {
	for i := range key {
		key[i] = 0
	}
	runtime.KeepAlive(key)
}
```

internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,371 @@
...
+	groupDir := filepath.Join(te.tokensDir, group)
+	if err := os.MkdirAll(groupDir, 0o700); err != nil {
+		return fmt.Errorf("create token directory: %w", err)
+	}
+
+	// Write the encrypted file to disk first. Only proceed to update
+	// the keyring once the disk write succeeds, so a disk failure
+	// never leaves the keyring ahead of the on-disk state.
+	filePath := filepath.Join(groupDir, fmt.Sprintf("token-%s.json.enc", pluginName))
+	if err := os.WriteFile(filePath, data, 0o600); err != nil {
```
sergio-correia: `os.WriteFile` truncates and rewrites in-place. If the process gets killed mid-write (OOM, power loss), the file is left truncated or empty — and this file holds the refresh token, which is the only way to get new access tokens without re-authenticating.

The codebase already uses atomic temp+rename in two other places: `migration_state.go:198-222` and `auth/oauth2.go:277-299`. Would be good to use the same pattern here — `os.CreateTemp` in the same directory, write, `Sync()`, then `os.Rename`.

There's also a subtler issue: if the disk write succeeds but the subsequent keyring write at line 108 fails, the old access token in the keyring is now paired with the new encrypted payload on disk (which may contain a rotated refresh token and new expiry). `LoadToken` would then construct a token where `Valid()` returns true (new expiry) but the access token is stale/revoked, preventing the refresh flow from kicking in.

internal/credentials/backup.go
```diff
@@ -0,0 +1,404 @@
...
+			groupCreds[key] = value
+		}
+		payload.CredentialGroups[group] = groupCreds
+	}
+
+	return payload, nil
+}
+
+const (
+	maxKDFMemory     = 4 * 1024 * 1024 // 4 GB in KB
```
sergio-correia: The validation bounds here are very generous compared to the production defaults:

| Param | Default | Max allowed | Ratio |
|-------|---------|-------------|-------|
| Memory | 64 MB | 4 GB | 64x |
| Iterations | 3 | 100 | 33x |
| Parallelism | 4 | 255 (uint8) | 63x |

A crafted backup file with `memory: 4194304` and `iterations: 100` would cause the victim's machine to allocate 4 GB and churn through Argon2 before GCM even gets to reject the ciphertext. In a container with a 1-2 GB limit, that's an instant OOM kill.

Please tighten these to something like 256 MB / 20 iterations / 16 parallelism, and add a parallelism check (currently unbounded). Also, `validateKDFParams` has no test coverage right now — boundary tests for this security validation would be valuable.

cmd/wtmcpctl/vault.go
```diff
@@ -168,6 +183,37 @@ func runVaultView(_ *cobra.Command, args []string) error {
...
+	}
+}
+
+// groupFromPath extracts the credential group name from a file path.
+// e.g., "env.d/github.env" → "github", "credentials/jira/token.json" → "jira"
+// For files nested under a named directory (e.g. credentials/jira/token.json),
+// the parent directory name is used as the group. For flat files (e.g.
+// env.d/github.env), the stem is used after stripping .enc and the remaining
+// extension.
+func groupFromPath(path string) string {
```
sergio-correia, line 210: The doc comment says `"env.d/github.env" -> "github"` but the implementation returns `"env.d"`. Trace: `filepath.Dir("env.d/github.env")` = `"env.d"`, `filepath.Base("env.d")` = `"env.d"`, which is not `"."`, `"/"`, or `".."`, so the function returns `"env.d"`. The flat-file stem path is never reached for any file under a named subdirectory.

This causes an AAD mismatch with the runtime decryption path. The CLI builds AAD with `group="env.d"`, but `config/env.go` derives group via `strings.TrimSuffix(filepath.Base(path), ".env")` which yields `"github"`. AES-256-GCM authentication fails because the AAD doesn't match. All `$WTMCP_VAULT` encrypted env.d files would fail to load at runtime.

CLI decrypt masks this because it uses the same `groupFromPath`, so round-trip testing through the CLI alone passes.

Cleanest fix is probably to pass the group name explicitly from the caller instead of inferring it from the path.

sergio-correia (commented):
Found some issues to address before merging, mostly around non-atomic multi-step writes, inconsistent secret zeroing, and a `groupFromPath` bug that breaks AAD binding between the CLI and runtime paths.

internal/secrets/decryptor.go
```diff
@@ -0,0 +1,229 @@
...
+	defer d.mu.Unlock()
+
+	if d.passwordCached {
+		pw := make([]byte, len(d.cachedPassword))
+		copy(pw, d.cachedPassword)
+		return pw, nil
+	}
+
+	// Try pinentry first (preferred for new vault format).
+	if d.pinentryClient != nil {
```
sergio-correia, line 190: When pinentry is present and the user cancels the password dialog (or pinentry crashes/times out), the error is silently swallowed and execution falls through to the Ansible resolver. If that also fails, the final error says "no vault password available -- install pinentry" — misleading because pinentry is installed and was used.

More importantly, if the fallback resolver succeeds with a stale password from an env var the user no longer trusts, the user's explicit cancellation was silently overridden.

This connects to a broader pattern: `Service.GetAll` (service.go:341) silently swallows env.d errors, and at line 351 silently swallows non-"not found" keyring errors. `pluginVars` (manager.go:1195) silently falls back to legacy env.d values for migrated groups when the keyring is temporarily unavailable. Combined, a temporary keyring outage silently reverts plugins to stale env.d credentials with zero logging.

Would be good to at least distinguish user cancellation from other pinentry errors here — cancellation should not fall through.

internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	// Combine ciphertext + tag for GCM Open.
+	sealed := make([]byte, 0, len(ciphertext)+len(tag))
+	sealed = append(sealed, ciphertext...)
+	sealed = append(sealed, tag...)
+
+	// Decrypt with AAD to verify group/plugin binding.
+	aad := []byte(group + "\x00" + pluginName)
+	plaintext, err := gcm.Open(nil, nonce, sealed, aad)
```
sergio-correia, line 392: The base64-decoded nonce is passed directly to `gcm.Open` without checking its length against `gcm.NonceSize()`. Go's `cipher.AEAD.Open()` panics on wrong nonce length — so a corrupted or tampered `.json.enc` file with a truncated IV field crashes the server process.

The encrypt path correctly uses `gcm.NonceSize()` to generate the nonce, but the decrypt path trusts whatever comes out of the base64 decode. Same issue in `backup.go:391` (`decryptBackupPayload`).

```suggestion
	nonce, err := base64.StdEncoding.DecodeString(encrypted.IV)
	if err != nil {
		return fmt.Errorf("decode IV: %w", err)
	}
	if len(nonce) != gcm.NonceSize() {
		return fmt.Errorf("invalid IV length %d (expected %d)", len(nonce), gcm.NonceSize())
	}
```

internal/credentials/backup.go
```diff
@@ -0,0 +1,404 @@
...
+
+	// Marshal the payload to JSON.
+	plaintext, err := json.Marshal(payload)
+	if err != nil {
+		return nil, fmt.Errorf("marshal payload: %w", err)
+	}
+	defer zeroBackupKey(plaintext)
+
+	// Encrypt with GCM. The sealed output is ciphertext + tag.
+	sealed := gcm.Seal(nil, nonce, plaintext, nil)
```
sergio-correia: Both `encryptBackupPayload` and `decryptBackupPayload` pass `nil` as AAD to GCM. The backup JSON envelope has `version`, `created_at`, `kdf`, `kdf_params` fields outside the authentication envelope — none of them are bound to the ciphertext.

Contrast with `token_encryption.go:307` which correctly uses `group + "\x00" + pluginName` as AAD, and `wtmcpvault/password.go:69` which also binds AAD. The backup path is the odd one out.

While tampering with `salt` or `kdf_params` is self-defeating (produces a different derived key, so GCM fails), it fails *after* the expensive Argon2 computation runs. An attacker who modifies `kdf_params` to extreme values (see the `maxKDFMemory` comment) gets the resource exhaustion before GCM rejects. Binding the params as AAD would cause rejection before `deriveKey` runs.

Something like:
```go
aad := []byte(fmt.Sprintf("wtmcp-backup-v%d;%s;m%d;t%d;p%d",
    1, "argon2id", argon2Memory, argon2Iterations, argon2Parallelism))
```

internal/secrets/decryptor.go
```diff
@@ -0,0 +1,229 @@
...
+	case wtmcpvault.BackendFIDO2:
+		return nil, fmt.Errorf("FIDO2 backend not yet implemented")
+	case wtmcpvault.BackendPKCS11:
+		return nil, fmt.Errorf("PKCS#11 backend not yet implemented")
+	default:
+		return nil, fmt.Errorf("unknown backend: %s", header.Backend)
+	}
+}
+
+func (d *MultiDecryptor) decryptWtmcpPassword(data []byte, headerLine, context string) ([]byte, error) {
```
sergio-correia, line 165: `resolveWtmcpPassword()` returns a fresh copy of the cached password (line 182-183), but `decryptWtmcpPassword` never zeroes that copy after use. Each call leaks one unzeroed password into the Go heap.

The Ansible vault path at line 136 handles this correctly with `defer vault.ZeroBytes(password)`.

```suggestion
func (d *MultiDecryptor) decryptWtmcpPassword(data []byte, headerLine, context string) ([]byte, error) {
	password, err := d.resolveWtmcpPassword()
	if err != nil {
		return nil, err
	}
	defer zeroBytes(password)

	aad := wtmcpvault.BuildAAD(headerLine, context)
	return wtmcpvault.DecryptPassword(data, password, aad)
}
```

internal/credentials/service.go
```diff
@@ -0,0 +1,358 @@
...
+// GroupInfo holds summary information about a credential group.
+type GroupInfo struct {
+	Name        string
+	Migrated    bool
+	EnvDKeys    []string
+	KeyringKeys []string
+}
+
+// GetGroupInfo returns detailed information about a credential group.
+func (s *Service) GetGroupInfo(group string) GroupInfo {
```
sergio-correia, line 297: Every other read method on `Service` (`Get`, `GetAll`, `ListGroups`, `CreateBackup`) acquires `s.mu.RLock()` before accessing `migrationState` or `envdStore`. `GetGroupInfo` does not.

A concurrent `RestoreBackup` or `Delete` (which hold `s.mu.Lock()`) can run simultaneously, producing a torn read — e.g., `Migrated: false` with `KeyringKeys` populated from a restore that's mid-flight.

```suggestion
func (s *Service) GetGroupInfo(group string) GroupInfo {
	s.mu.RLock()
	defer s.mu.RUnlock()

	info := GroupInfo{
		Name:     group,
		Migrated: s.migrationState.IsMigrated(group),
	}
```

internal/credentials/token_encryption.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	// Reconstruct the oauth2.Token.
+	token := &oauth2.Token{
+		AccessToken:  accessToken,
+		RefreshToken: payload.RefreshToken,
+		TokenType:    payload.TokenType,
+	}
+
+	if accessToken != "" && payload.Expiry != "" {
```
sergio-correia: When the access token is missing from the keyring (e.g., after a backup restore) but `payload.Expiry` is non-empty on disk, the expiry is never parsed because it's gated on `accessToken != ""`. The token gets zero expiry, and `oauth2.Token.Valid()` returns true for zero expiry (treats it as "never expires"). So if a subsequent code path populates the access token from another source, no refresh is ever triggered despite potentially stale credentials.

Setting the expiry to a past time when `accessToken` is empty would make the intent explicit:
```go
if accessToken == "" {
	token.Expiry = time.Now().Add(-time.Minute)
}
```

internal/credentials/backup.go
```diff
@@ -0,0 +1,404 @@
...
+func (s *Service) RestoreBackup(backupPath string, password []byte, mergeMode bool) error {
+	if len(password) == 0 {
+		return ErrEmptyPassword
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// Read the backup file.
+	data, err := os.ReadFile(backupPath) //nolint:gosec // user-provided path
```
sergio-correia, line 173: No size check before `os.ReadFile`. A multi-GB file (accidentally pointed at the wrong path, or crafted) gets read entirely into memory. The `wtmcpvault` password module already guards against this at `password.go:92` with `maxFileSize`.

```suggestion
func (s *Service) RestoreBackup(backupPath string, password []byte, mergeMode bool) error {
	if len(password) == 0 {
		return fmt.Errorf("empty password")
	}

	info, err := os.Stat(backupPath)
	if err != nil {
		return fmt.Errorf("stat backup file: %w", err)
	}
	if info.Size() > 10<<20 {
		return fmt.Errorf("backup file too large: %d bytes (max 10 MB)", info.Size())
	}

	data, err := os.ReadFile(backupPath)
```

internal/credentials/service.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// Start with env.d values (always available for legacy).
+	if envdVars, err := s.envdStore.GetAll(group); err == nil {
+		for k, v := range envdVars {
+			vars[k] = v
+		}
+	}
+
+	// Overlay keyring values for migrated groups.
+	if s.migrationState.IsMigrated(group) {
```
sergio-correia: `GetAll` only overlays keyring keys when `IsMigrated(group)` is true. But a user can store individual keys in the keyring via `wtmcpctl credentials set jira JIRA_TOKEN xxx --keyring` without migrating the whole group.

`Get("jira", "JIRA_TOKEN")` finds the key correctly through `getLegacyCredential`'s keyring fallback (line 127). But `GetAll("jira")` skips the keyring overlay because `IsMigrated("jira")` is false. Since `pluginVars()` calls `GetAll`, the plugin never receives `JIRA_TOKEN` in its template variables and fails to authenticate — with no clear error explaining why.

Should check `migrationState.GetGroupKeys(group)` and overlay those keys regardless of migration status.

internal/credentials/backup.go
rjeffman: Added rollback tracking to RestoreBackup. Written keys are tracked in a []restoredKey slice; on failure, all previously written keys are deleted from the keyring. Rollback errors are logged (not silenced) so the user knows manual cleanup may be needed. Also noted and preserved the existing rollback pattern in runCredMigrate.

internal/credentials/token_encryption.go
rjeffman: Replaced os.WriteFile with atomic temp+rename pattern: os.CreateTemp in the same directory, write, Sync(), Chmod(0600), then os.Rename. Crash mid-write leaves only the temp file — the previous .json.enc is untouched until rename succeeds.

internal/credentials/backup.go
rjeffman: Tightened bounds from 4GB/100/unbounded to 256MB (maxKDFMemory = 256*1024), 20 iterations (maxKDFIterations = 20), and 16 parallelism (maxKDFParallelism = 16). Added parallelism upper-bound check. Added TestValidateKDFParams table-driven test covering all boundary conditions and a TestRestoreBackup_RejectsOversizedFile test.

cmd/wtmcpctl/vault.go
rjeffman: Fixed groupFromPath to match the runtime derivation in config/env.go. For .env files, the stem is always used as the group name (e.g. env.d/github.env → "github"), regardless of the parent directory name. This ensures the AAD matches between CLI encrypt/decrypt and runtime decryption.

internal/secrets/decryptor.go
rjeffman: Added pinentry.ErrCancelled sentinel error (detected via GPG error code 83886179). resolveWtmcpPassword now returns the cancellation error immediately without falling through to the Ansible resolver. Other pinentry errors (timeout, crash) still fall through to preserve backward compatibility.

internal/credentials/token_encryption.go
rjeffman: Added len(nonce) != gcm.NonceSize() check after base64 decode in both decryptPayload (token_encryption.go) and decryptBackupPayload (backup.go). Returns a descriptive error instead of panicking on truncated/corrupted IV fields.

internal/credentials/backup.go
rjeffman: Added AAD to both encryptBackupPayload and decryptBackupPayload. The AAD string format is "wtmcp-backup-v{version};{kdf};m{memory};t{iterations};p{parallelism}", binding the backup envelope metadata to the ciphertext. Combined with the tightened KDF bounds, tampering with params is rejected by both validation and GCM authentication.

internal/secrets/decryptor.go
rjeffman: Added defer zeroBytes(password) to decryptWtmcpPassword, matching the Ansible vault path which already has defer vault.ZeroBytes(password).

internal/credentials/service.go
rjeffman: Added s.mu.RLock() / defer s.mu.RUnlock() to GetGroupInfo, consistent with every other read method on Service.

internal/credentials/token_encryption.go
rjeffman: When accessToken is empty (e.g. after backup restore), token.Expiry is now set to time.Now().Add(-time.Minute) so oauth2.Token.Valid() returns false and the refresh flow kicks in automatically.

internal/credentials/backup.go
rjeffman: Added os.Stat + size check before os.ReadFile in RestoreBackup, rejecting files larger than 10 MB (maxBackupFileSize). Added TestRestoreBackup_RejectsOversizedFile test.

internal/credentials/service.go
rjeffman: Changed GetAll to always overlay keyring keys from migrationState.GetGroupKeys(group), not just when IsMigrated(group) is true. This ensures keys stored via wtmcpctl credentials set --keyring before full migration are visible to pluginVars().

sergio-correia (approved):
Thanks!

PR #31, opened 2026-05-13
drive: add write tools and safety controls

walac:
The Google Drive plugin previously only supported read operations. This change introduces write capabilities by adding upload, rename, move, copy, soft-delete, and folder creation tools, expanding the plugin's utility for file management.

To support these operations, the OAuth scope is upgraded to full read-write access. A lazy scope probing mechanism is implemented using GenerateIds on the first write attempt. If the token lacks the required scope, the tool returns an error guiding the user to re-authorize. This probe automatically resets upon encountering a 403 error to ensure subsequent attempts re-evaluate the scope.

Safety is prioritized across all new operations. All write tools default to a dry-run mode and are marked with write access requirements to trigger user confirmation. Upload paths are strictly confined to the user's home directory, with symlink resolution enforced on both the home directory and target paths to prevent directory traversal escapes. Deletions are implemented as recoverable soft-deletes to the trash. Comprehensive test coverage validates these safety mechanisms, scope gating, and parameter handling.

Assisted-by: Claude Code:claude-opus-4-6 [PAL]

### DIFF
--- a/plugins/google-drive/tools.go
+++ b/plugins/google-drive/tools.go
@@ -3,12 +3,19 @@ package main
 import (
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	htmltomarkdown "github.com/JohannesKaufmann/html-to-markdown/v2"
+	"google.golang.org/api/drive/v3"
 	googleapi "google.golang.org/api/googleapi"
 )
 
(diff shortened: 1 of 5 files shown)


plugins/google-drive/tools.go
```diff
@@ -505,3 +512,380 @@ func buildSearchQuery(text string, inNameOnly bool, mimeTypes, owners []string,
...
+	hasWriteScope    bool
+)
+
+func requireWriteScope() error {
+	writeScopeMu.Lock()
+	defer writeScopeMu.Unlock()
+
+	if !writeScopeProbed {
+		writeScopeProbed = true
+		hasWriteScope = true
```
sergio-correia: Here we set `hasWriteScope = true` before probing. If `GenerateIds` fails with a network error or 500, the result is permanently cached as "has scope", and subsequent calls skip the probe and fail with raw API errors instead of the re-auth message. Let's set `hasWriteScope = true` only after a successful probe, and reset `writeScopeProbed = false` on transient errors, so the next call retries.

sergio-correia (changes requested):
You need to update `context.md` for write tools. It currently says the plugin is read-only and `"cannot create, modify, or delete files."`. Please update it to reflect the write capabilities.

Consider also adding `httptest-based` tests for the non-dry-run paths of upload, rename,  copy, and delete, similar to `TestToolCreateFolderActual`.

plugins/google-drive/tools.go
walac: Fixed!

sergio-correia:
@walac: Hi! Change looks good, but now CI is complaning, could you take a look at it, please? I attempted to remove the 2nd commit as it was addressed outside this PR, but that was not enough.

sergio-correia (approved):
LGTM, thanks!

PR #32, opened 2026-05-13
feat: add Jenkins plugin with Kerberos/SAML SSO authentication

rjeffman:
## What's New
                                                                                                                                                                                
**Jenkins Plugin**
- Browse jobs, folders, and views with health scores
- Get build details, logs, and test results (JUnit)
- Analyze pipeline stages with per-stage logs
- Download and inspect artifacts
- Smart caching (5min TTL, permanent for completed builds)
- Automatic secret redaction (AWS keys, JWTs, passwords, connection strings)
                                                                                                               
**Kerberos/SAML Authentication**
- Dynamic SPN derivation from request hostname
- Automatic SAML POST binding detection and replay
- Per-request domain allowlisting for SSO redirects
- Enhanced config variable resolver: `|hostname` modifier, nested defaults
                                                                                                                  
**Tools:** 15 tools total (4 primary, 11 deferred)
- `jenkins_list_jobs`, `jenkins_get_build`, `jenkins_get_build_log`, `jenkins_whoami`
- `jenkins_get_pipeline_stages`, `jenkins_get_test_results`, `jenkins_list_artifacts`, etc.
                                                                                                  
## Changes
                                                                                                                                                                                   
- Add `internal/proxy/saml.go` - SAML SSO redirect handler 
- Add `internal/auth/kerberos_provider.go` - dynamic SPN support
- Add `plugins/jenkins/` - complete plugin with tests
- Update `golang.org/x/net` to fix HTTP/2 vulnerability (GO-2026-4918)

### DIFF
--- a/env.d/jenkins.env.example
+++ b/env.d/jenkins.env.example
@@ -1,0 +1,62 @@
+# Jenkins plugin configuration
+#
+# Copy this file to jenkins.env and uncomment the variables you need:
+#   cp jenkins.env.example jenkins.env
+#
+# For security, consider encrypting with Ansible Vault:
+#   ansible-vault encrypt jenkins.env
+
+# Jenkins instance URL (used as default if not specified in tool calls)
+#JENKINS_URL=https://jenkins.example.com
+
+# Auth type: auto, user-token, bearer-token, kerberos
+# "auto" automatically selects based on available credentials.
+#JENKINS_AUTH_TYPE=auto
+
+# --- Username + API Token (most common) ---
+# Your Jenkins username
+#JENKINS_USER=your-username
+
+# API token from Jenkins > User > Configure > API Token
+# Get your token: Click your username (top-right) > Configure > API Token > Add new token
+#JENKINS_TOKEN=your-api-token-from-jenkins
+
+# --- Bearer Token (OAuth2 proxy) ---
+# Uncomment if Jenkins is behind an OAuth2 proxy
+# Set JENKINS_AUTH_TYPE=bearer-token
+#JENKINS_TOKEN=your-bearer-token
+
+# --- Kerberos (enterprise) ---
+# Kerberos uses your existing ticket (kinit)
+# Set JENKINS_AUTH_TYPE=kerberos
+# No credentials needed - uses system Kerberos ticket
+
+# --- Example Configurations ---
+
+# Example 1: Standard Jenkins with API token
+# JENKINS_USER=jdoe
+# JENKINS_TOKEN=11234567890abcdef1234567890abcdef
+
+# Example 2: Jenkins behind OAuth2 proxy
+# JENKINS_AUTH_TYPE=bearer-token
+# JENKINS_TOKEN=ya29.a0AfH6SMBx...
+
+# Example 3: Enterprise Jenkins with Kerberos
+# JENKINS_AUTH_TYPE=kerberos
+# (No credentials needed, uses kinit ticket)
+
+# --- Security Notes ---
+#
+# 1. Use Ansible Vault to encrypt this file:
+#    ansible-vault encrypt ~/.config/wtmcp/env.d/jenkins.env
+#
+# 2. Prefer API tokens over passwords:
+#    - API tokens can be scoped and revoked independently
+#    - They don't grant Jenkins UI login access
+#
+# 3. For Kerberos deployments:
+#    - No credentials stored in this file
+#    - Uses your kinit ticket (most secure)
+#
+# 4. File permissions:
+#    chmod 600 ~/.config/wtmcp/env.d/jenkins.env
--- a/plugins/jenkins/handler.py
+++ b/plugins/jenkins/handler.py
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+"""Jenkins plugin handler.
+
+Persistent handler with init/shutdown lifecycle. Auth and SAML SSO
+are handled by the Go core proxy. Cache is provided by the core
+cache service.
+"""
+
+import json
+import sys
+from urllib.parse import urlparse
+
+_next_id = 0
+
+# Plugin state set during init
+config = {}
+jenkins_version = "unknown"
+
+# Runtime Jenkins URL (set per-request by tools)
+_jenkins_url = ""
+
+
+def _send(msg):
+    """Write a JSON message to stdout (core reads it)."""
+    print(json.dumps(msg, separators=(",", ":")), flush=True)
+
+
+def _gen_id(prefix="svc"):
+    """Generate unique ID for requests."""
+    global _next_id
+    _next_id += 1
+    return f"{prefix}-{_next_id}"
+
+
+def _recv():
+    """Read a JSON message from stdin (core writes it)."""
+    line = sys.stdin.readline()
+    if not line:
+        sys.exit(0)
+    return json.loads(line.strip())
+
+
+def log(msg):
+    """Log to stderr (captured by core)."""
+    print(f"jenkins: {msg}", file=sys.stderr, flush=True)
+
+
+def validate_jenkins_url(raw_url):
+    """Validate and normalize a Jenkins URL.
+
+    Returns the normalized URL or None if invalid.
+    """
+    if not raw_url or not isinstance(raw_url, str):
+        return None
+    raw_url = raw_url.strip().rstrip("/")
+    if not raw_url:
+        return None
+    parsed = urlparse(raw_url)
+    if parsed.scheme not in ("https", "http"):
+        return None
+    if not parsed.hostname:
+        return None
+    return raw_url
+
+
+def set_jenkins_url(raw_url):
+    """Set the module-level Jenkins URL for subsequent http() calls.
+
+    Returns an error dict if invalid, None on success.
+    """
+    global _jenkins_url
+    validated = validate_jenkins_url(raw_url)
+    if validated is None:
+        return {"error": "invalid_input", "detail": f"invalid jenkins_url: {raw_url!r}"}
+    _jenkins_url = validated
+    return None
+
+
+def http(method, path, query=None, body=None, headers=None, url=None):
+    """Make an HTTP request via the core proxy.
+
+    Returns (status, body, headers). Status 0 means transport error.
+    When url is provided, it overrides base_url + path.
+    Uses _jenkins_url to construct full URLs and register domains.
+    """
+    msg = {
+        "id": _gen_id("http"),
+        "type": "http_request",
+        "method": method,
+        "path": path,
+    }
+    if url:
+        msg["url"] = url
+    elif _jenkins_url:
+        msg["url"] = _jenkins_url + path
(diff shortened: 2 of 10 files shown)


plugins/jenkins/plugin.yaml
```diff
@@ -0,0 +1,392 @@
...
+      bearer-token:
+        type: bearer
+        token: "${JENKINS_TOKEN}"
+      kerberos:
+        type: kerberos/spnego
+        spnego_proactive: false
+  http:
+    allow_private_ips: true
+    allowed_domains:
+      - jenkins-csb-idmops-ci.dno.corp.redhat.com
```
sergio-correia: Replace this with `"${JENKINS_URL|hostname}"`

plugins/jenkins/plugin.yaml
```diff
@@ -0,0 +1,392 @@
...
+  cache:
+    enabled: true
+    namespace: jenkins
+    default_ttl: 300
+
+env:
+  - JENKINS_USER
+  - JENKINS_TOKEN
+  - JENKINS_AUTH_TYPE
```
sergio-correia, line 41: Add `JENKINS_URL` to the env list here.

env.d/jenkins.env.example
```diff
@@ -0,0 +1,59 @@
+# Jenkins plugin configuration
+#
+# Copy this file to jenkins.env and uncomment the variables you need:
+#   cp jenkins.env.example jenkins.env
+#
+# For security, consider encrypting with Ansible Vault:
+#   ansible-vault encrypt jenkins.env
+
+# Auth type: auto, user-token, bearer-token, kerberos
```
sergio-correia, line 12: Let's add a `#JENKINS_URL=https://jenkins.example.com` here.

decko:
I think you need to rebase your work with current changes on wtmcp core @rjeffman. Let me know so we can help with the review of Jenkins plugin again

env.d/jenkins.env.example
rjeffman: Fixed

plugins/jenkins/plugin.yaml
rjeffman: Fixed

plugins/jenkins/plugin.yaml
rjeffman: Fixed

sergio-correia (commented):
Looks good overall. I pointed out a few issues that should be addressed.

plugins/jenkins/tools_read.py
```diff
@@ -0,0 +1,1254 @@
...
+        params: dict with jenkins_url
+
+    Returns:
+        dict: User info with id, fullName, email
+    """
+    err = _setup_url(params)
+    if err:
+        return err
+
+    cache_key = "jenkins:whoami"
```
sergio-correia: Cache keys don't include the Jenkins URL, but the plugin accepts `jenkins_url` per-request. If someone queries two different Jenkins instances in the same session, the second call returns cached data from the first.

For example, `jenkins:whoami` is a static key shared across all instances. Same for `jenkins:jobs:{folder_hash}` when two servers have identically named folders (extremely common in enterprise).

Please add a URL-derived hash to every cache key. Something like:

```python
url_hash = _cache_key_hash(handler._jenkins_url)
cache_key = f"jenkins:{url_hash}:whoami"
```

A `_make_cache_key(*parts)` helper that auto-includes the URL hash would prevent regressions across all 15 tools.

plugins/jenkins/plugin.yaml
```diff
@@ -0,0 +1,379 @@
...
+  - JENKINS_URL
```
sergio-correia, line 42: The `JENKINS_URL` env var fallback in `_setup_url()` reads from `handler.config.get("JENKINS_URL")`, but there is no `config:` section in this file. The core only populates `handler.config` from `manifest.Config` plus internal fields (`_session_dir`, `_output_dir`).

Without this, every user who sets `JENKINS_URL` in `env.d/jenkins.env` and calls a tool without passing `jenkins_url` explicitly will get an `invalid jenkins_url: ''` error. The fallback advertised in the tool descriptions never works.

Please add after the `env:` block:

```yaml
config:
  jenkins_url: "${JENKINS_URL}"
```

Then update `tools_read.py:31` to `handler.config.get("jenkins_url", "")` (lowercase, matching the config key).

plugins/jenkins/tools_read.py
```diff
@@ -0,0 +1,1254 @@
...
+
+    if status not in (200, 201):
+        return {"error": f"HTTP {status}", "detail": str(body)}
+
+    auth_error = helpers.check_auth_redirect(status, body, headers)
+    if auth_error:
+        return auth_error
+
+    # Scrub secrets
+    scrubbed_log = helpers.scrub_log_content(body)
```
sergio-correia, line 1077: The wfapi/log endpoint returns JSON like `{"text": "<log content>", "length": N, "hasMore": false}`. The proxy parses this into a dict, so `body` here is a dict, not a string. `scrub_log_content()` returns `""` for non-string input, so this will always produce an empty log against a real Jenkins instance.

The tests mask this because they mock `body` as a plain string.

Please extract the text field before scrubbing:

```python
if isinstance(body, dict):
    log_text = body.get("text", "")
elif isinstance(body, str):
    log_text = body
else:
    log_text = ""
scrubbed_log = helpers.scrub_log_content(log_text)
```

plugins/jenkins/handler.py
```diff
@@ -0,0 +1,226 @@
...
+def cache_flush(namespace):
+    """Flush all cache entries for a namespace."""
+    msg = {
+        "id": _gen_id("cache"),
+        "type": "cache_flush",
+        "namespace": namespace,
+    }
+    _send(msg)
+    resp = _recv()
+    return resp.get("status") == "ok"
```
sergio-correia: The core returns `{"ok": true, "count": N}` for cache_flush, not `{"status": "ok"}`. This means `cache_flush()` always returns `False`, so `jenkins_flush_cache` always reports failure even though the cache was actually flushed.

Also, the `namespace` parameter is ignored by the core (it always uses the plugin's own namespace). The jira plugin's simpler pattern is the right reference here:

```python
def cache_flush():
    _send({"id": _gen_id("cache"), "type": "cache_flush"})
    resp = _recv()
    return resp.get("ok", False)
```

plugins/jenkins/tools_read.py
```diff
@@ -0,0 +1,1254 @@
...
+
+    result = {
+        "job_name": job_name,
+        "build_number": build_number,
+        "artifact_path": artifact_path,
+        "content": content,
+        "size_bytes": len(body),
+    }
+
+    handler.cache_set(cache_key, result, ttl=0)
```
sergio-correia: This caches permanently with no checks, but every other tool guards against two things before caching: (1) whether the build is still running, and (2) whether `build_number` is an alias like `lastBuild`.

Without the running-build check, partially written artifacts get permanently cached. Without the alias check, `lastBuild` artifacts get frozen to whatever build was current at first fetch.

Please add the same pattern used in `jenkins_get_build` and `jenkins_get_build_log`:

```python
build_path = f"/{helpers.job_path(job_name)}/{build_number}/api/json"
build_status, build_body, _ = handler.http("GET", build_path, query={"tree": "building"})
building = build_body.get("building", False) if build_status in (200, 201) else False
if not building:
    try:
        int(build_number)
        handler.cache_set(cache_key, result, ttl=0)
    except (ValueError, TypeError):
        pass
```

plugins/jenkins/helpers.py
```diff
@@ -0,0 +1,533 @@
...
+                failures.append(
+                    {
+                        "name": testcase.get("name", ""),
+                        "classname": testcase.get("classname", ""),
+                        "message": failure_msg,
+                        "trace": failure_trace,
+                    }
+                )
+
+    passed = total - failed - skipped
```
sergio-correia: The JUnit XML spec defines `errors` alongside `failures` on `<testsuite>`. Errors represent unexpected exceptions (distinct from assertion failures). This line computes `passed = total - failed - skipped` without accounting for errors.

A testsuite with `tests="10" failures="1" errors="3" skipped="0"` would report 9 passed when only 6 actually passed. The `<error>` child elements on testcases are also not scanned, so those test cases silently disappear from the results.

Please add `suite_errors = int(testsuite.get("errors", 0))` and include it: `passed = total - failed - errors - skipped`. Also scan for `<error>` elements alongside `<failure>`.

plugins/jenkins/tools_read.py
rjeffman: Fixed.

plugins/jenkins/tools_read.py
rjeffman: Fixed.

plugins/jenkins/tools_read.py
rjeffman: Fixed

plugins/jenkins/helpers.py
rjeffman: Fixed.

plugins/jenkins/handler.py
rjeffman: Fixed.

plugins/jenkins/plugin.yaml
rjeffman: Fixed

sergio-correia:
Please, remove the ci update commit, as it is causing a conflict. That has been addressed outside this PR.

rjeffman:
That was interesting... the commit didn't show with `git log` (I thought I had removed), but was "lying around". Removed.

plugins/jenkins/tools_read.py
```diff
@@ -0,0 +1,1286 @@
...
+        "skipped": parsed["skipped"],
+        "duration_seconds": parsed["duration_seconds"],
+        "failures": failures,
+        "total_failures": total_failures,
+        "building": building,
+    }
+
+    # Cache permanently if build complete
+    if not building:
+        handler.cache_set(cache_key, result, ttl=0)
```
sergio-correia: Three tools permanently cache results under the raw `build_number` key, even when it is an alias like `"lastBuild"` or `"lastFailedBuild"`. Once cached, subsequent calls with the same alias silently return stale data forever.

`jenkins_get_build` (line 240-250) and `jenkins_get_build_log` (line 363-370) already handle this correctly by either resolving the alias to a numeric key or skipping the cache. The same guard is missing here and in `jenkins_get_pipeline_stages` (line 1007-1009) and `jenkins_get_stage_log` (line 1108-1110).

```suggestion
    # Cache permanently if build complete and build_number is numeric
    if not building:
        try:
            int(build_number)
            handler.cache_set(cache_key, result, ttl=0)
        except (ValueError, TypeError):
            pass  # Alias - don't cache permanently
```

rjeffman: fixed

sergio-correia (approved):
LGTM, thanks

PR #33, opened 2026-05-19
feat(proxy): add saml_init for proactive SAML SSO login

decko:
## Summary

- Cherry-picks the SAML SSO core from #32 (3 commits by @rjeffman — deps, auth, security)
- Adds `saml_init` config option for proactive SAML login during plugin startup

## Problem

The reactive SAML SSO from #32 triggers on 401/403 + text/html responses. Some sites (e.g. Igloo-based platforms) never return 401/403 — they silently degrade to anonymous sessions with empty results. The only way to authenticate is to explicitly initiate the SAML redirect flow.

## Solution

New `saml_init` field in the auth config:

```yaml
services:
  auth:
    type: kerberos
    spnego_proactive: false
    saml_init: "/saml.redirect?relayState="
```

When set, the core proactively follows the SAML flow during plugin startup:
1. GET the `saml_init` URL
2. Parse SAMLRequest form, POST to IdP (SPNEGORoundTripper handles Kerberos)
3. Parse SAMLResponse, POST back to origin server
4. Cookie jar is authenticated before any tool calls

Handles two patterns:
- **Form-first**: init URL returns a SAMLRequest form directly (Igloo, etc.)
- **Redirect-first**: delegates to the existing `handleSAMLSSO` for meta-refresh redirects (Jenkins, etc.)

SAML init failure is **fatal** — the plugin won't load, preventing silent degradation to empty results.

## Changes

- `internal/plugin/manifest.go` — add `SAMLInit` field to `AuthServiceConfig`
- `internal/proxy/saml.go` — add `InitSAMLSession` + `followSAMLFormChain`
- `internal/proxy/saml_test.go` — 5 new tests (form-first, redirect-first, no content, blocked domain, absolute URL)
- `internal/plugin/manager.go` — call `InitSAMLSession` during `preparePlugin` if configured

## Test plan

- [x] All existing SAML tests pass (6/6)
- [x] New InitSAMLSession tests pass (5/5)
- [x] Full test suite passes (`go test ./...`)
- [ ] Live test against an Igloo-based site with `saml_init` configured


rjeffman:
@decko reviewed with some agents, and this is what they find and I agree we should at least discuss:

## BLOCKING Issues - Must Fix Before Merge                                                      
                                                                                               
1. isDomainAllowed Normalization Inconsistency (CRITICAL)                                    
                  
  File: internal/proxy/proxy.go, lines 694-696
  Consensus: All 4 agents agree this is critical

  Issue: isDomainAllowed uses strings.EqualFold while isDomainAllowedForSSO and all other
  domain validation uses domaincheck.Contains with IDNA normalization.

  Impact: Domains with trailing dots, Unicode characters, or punycode could bypass validation
  at the primary request gate while being accepted/rejected differently by other validation
  paths.

  Fix: Replace the strings.EqualFold loop with domaincheck.Contains:
```
  // Change this:
  for _, domain := range pa.AllowedDomains {
      if strings.EqualFold(reqHost, domain) {
          return true
      }
  }
  // To this:
  return domaincheck.Contains(pa.AllowedDomains, reqHost)
```
2. AddBaseDomains Missing Cap (MEDIUM - Security)

  File: internal/proxy/proxy.go, lines 175-193
  Security-engineer: Blocks merge. Other agents concur.

  Issue: AddBaseDomains doesn't enforce maxTotalDomains cap, while AddAllowedDomains does.
  A malicious plugin could flood base domains and bypass subdomain restrictions.

  Fix: Add the same cap check that exists in AddAllowedDomains:
```
  if len(pa.AllowedDomains) >= maxTotalDomains {
      log.Printf("[%s] WARNING: domain allowlist cap reached (%d domains), rejecting %q",
          pluginName, maxTotalDomains, d)
      continue
  }
```

## NON-BLOCKING Issues - Acceptable to Defer

3. HTML Parsing via Regex

  Consensus: Safe due to domain validation, but may fail against some real IdPs
  - Security-engineer: H3 (Warning) - "mitigated by domain validation"
  - Rogue-reviewer: Concedes it's safe security-wise, but will cause functional failures
  - Code-reviewer: Upgraded to "Critical for functionality" but safe security-wise

  Recommendation: Track as high-priority follow-up to replace with golang.org/x/net/html for
  better IdP compatibility. Not merge-blocking because domain validation provides independent
  security control.

## Minor issues

- context.Background() in InitSAMLSession - should use timeout (HIGH priority follow-up)

rjeffman:
LGTM.

PR #66, opened 2026-05-20
google-docs: reduce API calls for multi-table writes and add rate-limit retry

sergio-correia:
- Reduce API calls ~48% for multi-table documents by deferring cell population to chunked batches
- Add doWithRetry with automatic rate-limit recovery for all Google Docs API calls; raise default rate limits to 300/m
- Fix duplicate tab prefixes in formatted nested list items (- **bold** text)


decko (commented):
## Code Review — PR #66

Reviewed the full diff across all 4 files. Focused on correctness, safety, and failure modes. Skipping cosmetic and minor items.

---

### Major: `sleepFunc` global var is a test-time data race risk

**`plugins/google-docs/tools.go:73`**

`sleepFunc` is a mutable package-level `var` overridden by tests via save/restore pattern. The comment says *"Tests that override it must NOT use t.Parallel()"*, but nothing enforces this — Go's test runner won't prevent a future contributor from adding `t.Parallel()` to these subtests.

**Recommendation:** Pass `sleepFunc` as a parameter to `doWithRetry`, or use a struct with an injectable sleep field. This eliminates the race by design rather than by convention.

---

### Major: `parseRateLimitDelay` relies on error string format — fragile coupling

**`plugins/google-docs/tools.go:95-122`**

The function parses `[rate_limited]` and `retry after <duration>` from error `.Error()` strings. This couples the plugin to the exact error message format produced by `internal/proxy`. If the proxy error format changes (e.g., different wording, structured errors, or `errors.As` wrapping), retry silently degrades to a 1s fixed delay.

**Mitigating factor:** The fallback to 1s is safe — it won't break, just retries suboptimally.

**Recommendation:** Consider exporting a sentinel error type or interface from the proxy package (e.g. `RateLimitError` with a `RetryAfter() time.Duration` method) so the plugin can use `errors.As` instead of string parsing. This is a cross-package change but would be more robust long-term.

---

### Major: Partial batch failure in Phase 2 leaves document in inconsistent state

**`plugins/google-docs/tools.go:1727-1740`**

Phase 2 sends multiple `BatchUpdate` calls to populate table cells. If batch 2/3 fails:
- Batch 1 has already been applied (some tables populated)
- The document is left with a mix of populated and empty tables
- The error message `"populate table cells (batch 2/3): %w"` doesn't indicate which tables were affected

There's no rollback mechanism, and the caller has no way to know which tables are complete.

**Recommendation:** At minimum, include the table indices in the error message so the user/caller knows which tables need manual attention. Consider whether the function should return partial success info alongside the error.

---

### Major: Config backfill ignores intentional opt-out

**`internal/config/config.go:291-300`**

The backfill loop always adds Google API rate-limit defaults for any domain not explicitly present in `per_domain`. If a user sets:

```yaml
http:
  rate_limit:
    per_domain: {}
```

...intending to clear all per-domain overrides and use the global default (120/m) for everything, the backfill silently re-adds the Google 300/m entries. There's no way to opt out of the elevated Google limits without explicitly setting each domain to a lower value.

**Recommendation:** Either document this behavior clearly, or add a convention (e.g., setting a domain to `""` or `"default"` explicitly opts out of the built-in override).

---

### Notable (not blocking): Two-phase index stability looks correct

I traced through the Phase 1 / Phase 2 logic carefully. Phase 1 processes segments in forward order, always inserting at or after `currentIndex`, so previously-created table cell indices are never invalidated. Phase 2 processes tables in reverse document order via `buildCellBatches`, which correctly preserves the invariant that later insertions don't shift earlier cell positions.

The `convertMarkdownToRequests` finalIndex calculation with `tabsInserted` accounting is well-tested (22 test cases covering plain text, dates, code blocks, nested lists with formatting, multi-byte UTF-8, etc.). The `isContinuation` detection for avoiding duplicate tab prefixes in formatted nested list items is a subtle fix — the logic correctly checks the previous segment's properties after `mergeSegments` has been applied.

---

### Test coverage assessment

**Well covered:**
- `doWithRetry` — success, non-rate-limit error, retry-then-succeed, exhausted retries (4 tests)
- `parseRateLimitDelay` — nil, non-rate-limit, various durations, clamping, parsing failures (10 tests)
- `buildCellBatches` — empty, small, multi-table, splitting, reverse order, maxReqs=0 (7 tests)
- `convertMarkdownToRequests` finalIndex — extensive table-driven tests (22 cases)
- Config per-domain merge and user override (3 tests)

**Gap:** `insertMarkdownWithTables` itself has no unit test — it requires a live Google Docs API service. The two-phase logic is tested indirectly through its sub-functions (`convertMarkdownToRequests`, `buildCellBatches`, `convertTableToRequests`). This is an acceptable tradeoff given the external dependency, but worth noting.

---

**Summary:** No blockers. Three major design concerns (global `sleepFunc`, error string parsing, partial batch failure) and one config UX issue. The core algorithmic changes (two-phase insertion, finalIndex tracking, tab fix) are sound and well-tested.

decko:
## MCP Architecture Review

### CRITICAL: Retry logic belongs in the proxy, not the plugin

The google-docs plugin parses error *strings* from `ProxyTransport.RoundTrip` (`fmt.Errorf("proxy transport: [%s] %s", resp.Error.Code, resp.Error.Message)`) to detect rate limiting. This is a fragile coupling across the plugin-proxy boundary.

**The proxy already has retry infrastructure for this.** It has exponential backoff, `Retry-After` header parsing, context-aware sleep (`sleepContext`), and the configurable retry loop at `proxy.go:368-416`. The proxy *knows* the retry-after delay — it computes it via `rateLimiter.Allow(domain)` — but instead of sleeping and retrying, it immediately returns an error to the plugin.

**Problems with retry at the plugin layer:**

1. **String parsing contract** — `parseRateLimitDelay` depends on the exact format `"[rate_limited]"` and `"retry after "`. If anyone changes the format string in `proxy.go:347`, the plugin silently stops retrying. No compile-time or protocol-level contract enforces this.

2. **Duplicated retry logic** — The proxy has a configurable retry loop (max retries, backoff, retry-on status codes). The plugin creates a second, independent retry system with different constants (`maxRateLimitRetries=3`, `maxRetryDelay=30s`). These will diverge over time.

3. **Cross-plugin gap** — google-calendar, google-drive, and google-gmail all hit the same rate limiter (same domains, same config) but get zero retry benefit. If retry belongs at the plugin layer, it needs to be in a shared package (e.g., `internal/google/`), not per-plugin.

4. **Not context-aware** — `doWithRetry` uses `time.Sleep`, not a context-aware sleep. The proxy already has `sleepContext` for this. A tool call that retries 3x with 30s delays (worst case 90s) will block beyond the 60s `tool_call_timeout` — and cancellation from the MCP client or timeout does **not** interrupt the sleep.

**Recommendation:** Add the synthetic rate-limit code to the proxy's `retryOn` list and have the proxy sleep for the duration it already computed from `rateLimiter.Allow()`. This gives all plugins automatic retry with zero plugin-side code changes, uses context-aware sleep, and eliminates the string-parsing coupling.

If the decision is to keep retry in the plugin layer as a pragmatic short-term choice, then at minimum the sleep must be context-aware, and `doWithRetry` should be extracted to a shared package so the other Google plugins benefit.

rjeffman:
In my human, limited review, changes LGTM.

sergio-correia:
@decko: moved retry to the proxy and addressed the other points raised.

The rate-limit check in Execute now sleeps for the duration from Allow() and retries (up to 2 times, 15s cap per wait) instead of returning an error immediately. Uses the existing sleepContext for context-aware cancellation. All plugins get automatic rate-limit recovery with zero plugin-side code.

`doWithRetry`, `parseRateLimitDelay`, and `sleepFunc` are deleted from google-docs. The string-parsing coupling, cross-plugin gap, and `time.Sleep` blocking are all gone.

Also fixed a pre-existing issue in `Allow()` where the per-key token was consumed but not returned when the global limiter rejected. The retry loop would have amplified that.

decko:
## Specialist Review — Post May 26 Update

Retry-to-proxy migration, ratelimit bug fix, two-phase batching refactor, and finalIndex tracking all look correct. Previous review concerns are addressed. Three items remain.

---

### 1. Dimension mismatch validation removed (regression)

**`plugins/google-docs/tools.go` — Phase 1 table lookup**

The old code explicitly checked `actualRows < expectedRows` and returned an error. The new code removes that check and relies on `collectTableCellRequests` silently skipping mismatched cells via `continue`. If the API returns fewer rows than requested, cells are quietly dropped with no error or warning.

**Recommendation:** Re-add the check before storing the `tableRecord`:

```go
if len(elem.Table.TableRows) < expectedRows {
    return nil, fmt.Errorf("table %d dimension mismatch: requested %d rows, API returned %d",
        tableIdx, expectedRows, len(elem.Table.TableRows))
}
```

### 2. Missing test for the ratelimit bug fix

**`internal/ratelimit/ratelimit.go` — per-key cancel on global reject**

The fix itself is correct and complete. But there's no test exercising the actual scenario: per-key allows (consumes token), global rejects, per-key token must be returned. `TestPerKeyBeforeGlobal` only tests per-key rejecting *before* global is checked.

**Suggested test:** Registry with `per_key="100/s"`, `global="1/s"`. First request passes both. Second request: per-key allows, global rejects. Verify per-key budget isn't silently drained.

### 3. Phase 2 batch error could include table indices

**`plugins/google-docs/tools.go` — Phase 2 batch loop**

Current error: `"populate table cells (batch 2/4, 3 tables): ..."`. Adding which table indices the failing batch covers would help diagnose partial-state documents. Not blocking — already better than the old per-table message.

---

Everything else LGTM. The ratelimit fix is sound, proxy retry is clean and context-aware, finalIndex arithmetic is well-tested (22 cases), config backfill is intentional and user-overridable, and the `highlighter/config.go` + `fileio_test.go` changes are pure refactors.

sergio-correia:
Thanks for the review! All three addressed.

**1. Dimension mismatch validation** - re-added. Went a step further though: the old search required exact row+column match to find the table element, so a mismatch gave a generic "table not found" error. Now the search finds the first table by position, then validates dimensions separately with a descriptive error (`"table 0 dimension mismatch: requested 3x2, API returned 2x2"`).

**2. Ratelimit test** - this one turned out to be deeper than expected. The sandboxed review caught that `rate.Reservation.Cancel()` is a no-op for zero-delay reservations (confirmed by reading `x/time/rate` source, `CancelAt` returns early when `timeToAct.Before(now)`). So the Cancel-based fix was non-functional, and the suggested test with burst=100 wouldn't have caught it either.

The fix: reversed `Allow()` to check global first, then per-key. When global rejects, per-key is never touched, no Cancel needed. The inverse trade-off (global token leaked when per-key rejects) is acceptable since global capacity is ~5x per-key. `TestGlobalRejectSkipsPerKey` validates this with a tight per-key burst (2/s) and a sleep+replenish cycle.

**3. Batch error indices** - done. Also went further here: `buildCellBatches` now aligns batch boundaries to cell boundaries (never splits a cell's InsertText from its UpdateTextStyle across batches), and `convertMarkdownToRequests` takes a `stripTrailingNewline` parameter so mid-document text flushes preserve trailing paragraph breaks.

decko:
## Specialist Review — Post May 29 Update

All three items from the previous round are correctly addressed. Four specialist agents reviewed the current diff: code reviewer, silent failure hunter, test coverage analyzer, and type design analyzer. Combined findings below.

---

### Verdict on the Three Addressed Items

1. **Dimension mismatch validation** — ✅ Correct. Now checks both rows AND columns (`actualRows != expectedRows || actualCols != expectedCols`). Search-then-validate split gives clear "not found" vs "dimension mismatch" errors with descriptive `%dx%d` format. Good improvement over the original.

2. **Global-before-per-key ratelimit ordering** — ✅ Correct. The discovery that `rate.Reservation.Cancel()` is a no-op for zero-delay reservations is well-documented in the code comments. The new ordering avoids the real bug (per-key token leak on global reject). The trade-off (global token leak when per-key rejects) is acceptable at 600/min vs 120-300/min capacity. `TestGlobalRejectSkipsPerKey` validates the invariant correctly.

3. **Batch error indices + cell-aligned batching + `stripTrailingNewline`** — ✅ Correct. `buildCellBatches` correctly processes tables in reverse document order. `finalIndex` arithmetic verified against multiple test cases including nested lists with formatting, multi-byte UTF-8, and date chips. `tabsInserted` tracking and `isContinuation` detection are both sound.

---

### New Findings

#### 1. Phase 2 partial failure leaves document in ambiguous state

**`plugins/google-docs/tools.go` — Phase 2 batch loop**

If Phase 2 fails mid-batch (e.g., batch 2/3 gets a transient API error), the document has:
- All table structures created (Phase 1 committed)
- Some cells populated (earlier batches committed)
- Some cells empty (failing + later batches never ran)

The error message lists ALL table indices, not which ones were populated vs empty. The caller can't distinguish "no tables populated" from "tables 0-1 done, table 2 partial."

The old per-table approach left tables 0..N-1 fully populated on failure at table N — a cleaner partial state.

**Suggestion:** Include batch progress in the error: `"populate table cells (batch 2/3 failed; 1/3 committed): ..."`. Not blocking, but improves debuggability for multi-table documents.

#### 2. `currentIndex = finalIdx` applied even when BatchUpdate is skipped

**`plugins/google-docs/tools.go` — Phase 1 text flush (lines ~778-791)**

When `pendingSegments` is non-empty but `convertMarkdownToRequests` returns zero requests, the BatchUpdate is skipped but `currentIndex = finalIdx` is still set. Currently unlikely (non-empty segments should always produce requests), but the code structure allows it — a future refactoring could silently produce a wrong `currentIndex`.

**Suggestion:** Move `currentIndex = finalIdx` inside the `if len(textReqs) > 0` block.

#### 3. `buildCellBatches` — single oversized cell group exceeds `maxReqs`

**`plugins/google-docs/tools.go` — `buildCellBatches`**

A single heavily-formatted cell producing more requests than `maxReqs` lands in one batch uncapped, because the split condition `len(current) > 0 && ...` is false when `current` is empty. The comment documents this is a "conservative heuristic" (API limits payload size, not count), so this is by design — but it's undocumented and untested.

**Suggestion:** Either add a comment noting cells are atomic (cannot be split across batches), or add a test case documenting this behavior. Not blocking.

---

### Test Coverage Notes

- **Strong overall.** 7 proxy rate-limit tests, 7 `buildCellBatches` sub-tests, 22+4 `finalIndex` cases, 3 config backfill tests. Clean dependency injection via `rateLimitChecker` interface and `sleepFn`.
- **`TestGlobalRejectSkipsPerKey` uses `time.Sleep(1100ms)`** — timing-dependent but margin is generous (10% buffer) and `rate.Limiter` uses monotonic time. Acceptable, but if it flakes, bump to 1500ms.
- **Minor gap:** `stripTrailingNewline=false` only has 4 test cases. The production path that uses it (Phase 1 text flush) could benefit from date/person chip cases, but the strip logic only affects the last segment's trailing `\n`, so risk is low.

### Type Design Notes

- **`rateLimitChecker` interface** — clean extraction, follows Go convention. Minor: `SetRateLimiter` still accepts `*ratelimit.Registry` instead of the interface, so tests must assign the field directly. One-line fix for consistency, not blocking.
- **`tableRecord` struct** — pragmatic data carrier. Would benefit from a constructor validating `len(apiRows) == len(parsedRows)`, since `collectTableCellRequests` silently skips mismatched rows via `continue`. Low severity given the dimension check at construction site.

---

### Summary

All previous concerns resolved. Three new items found — #1 (partial-failure error message) is the most impactful for production debuggability, #2 and #3 are defensive improvements. None are blocking. LGTM with optional improvements.

sergio-correia:
Thanks for the follow-up. All three addressed.

**1. Phase 2 partial failure** - the error message now includes batch progress: `"batch 2/3 failed; 1/3 committed, tables [0,1,2]"` so it's clear how much was committed vs lost. Also added a comment documenting that cell groups are atomic (never split across batches) and that a single oversized cell gets its own batch by design.

**2. `currentIndex = finalIdx` when BatchUpdate is skipped** - moved inside the `if len(textReqs) > 0` block so it only applies when the BatchUpdate actually executes. Also added a runtime sanity check after table creation that compares the predicted index against the API-reported `elem.StartIndex`, logging a warning if they diverge. This acts as a self-healing canary for any future drift in the index arithmetic.

**3. Oversized cell group** - added a comment explaining the design: cells are atomic and cannot be split across batches, so a single cell exceeding `maxReqs` gets its own batch. The API limits payload size (~10 MB), not request count, so this is safe in practice.

While working on these I also noticed and fixed two other things:

- `SetRateLimiter(nil)` had a Go typed-nil interface gotcha: assigning a nil `*ratelimit.Registry` to the `rateLimitChecker` interface field produced a non-nil interface, bypassing the nil check in the proxy. Added a nil guard in the setter.
- `TestGlobalRejectSkipsPerKey` would pass under both the old and new check orders because the 1.1s sleep let both limiters replenish. Added an immediate third `Allow()` call before the sleep that distinguishes the two orderings, making the test actually verify what it claims.

decko (approved):
All previous review concerns resolved across 4 rounds. The two bonus fixes (typed-nil interface guard, test distinguishability improvement) are correct. One minor nit: the `SetRateLimiter` comment says "A nil registry disables rate limiting" but the implementation ignores nil when a limiter is already set — phrasing could be tightened, but not blocking. LGTM.

sergio-correia:
Thanks for the reviews!

PR #67, opened 2026-05-20
feat(wtmcpctl): implement check command (#55)

### LINKED ISSUE #55: Implement wtmcpctl check command
## Description

`wtmcp check` exists (`cmd/wtmcp/main.go:65`) and runs plugin discovery, config validation, vault status, and env group checking — all outputting to stdout (not MCP protocol). Users need a way to validate their setup from the human-facing CLI tool (`wtmcpctl`), not the server binary.

Implement `wtmcpctl check` that reuses the same `plugin.Discover()` function and produces identical output to `wtmcp check`. This is the user's pre-flight verification step before their AI client tries to launch the server.

## Implementation Approach

1. Create `cmd/wtmcpctl/check.go` with a new `checkCmd` cobra command
2. The command should call `plugin.Discover()` with the same `DiscoveryOptions` pattern used in `cmd/wtmcp/main.go:320-398`
3. Print the same diagnostic output: version, workdir, plugin mode, vault status, env groups, plugin search path, discovered plugins with tool counts
4. Register it in `main.go` via `rootCmd.AddCommand(..., checkCmd)`
5. The command should respect the `--workdir` global flag (already handled by `PersistentPreRunE`)

Reference the existing `runCheck()` function in `cmd/wtmcp/main.go:320-398` for the exact output format. The `wtmcpctl` version can import and use `plugin.Discover()` and `plugin.DiscoveryOptions` from `internal/plugin/discovery.go`. The existing `cmd/wtmcpctl/common.go` has the `resolveWorkdir()` helper already.

## Acceptance Criteria

- [ ] `wtmcpctl check` runs without error and produces diagnostic output
- [ ] Output includes: version, workdir path, plugin mode, vault password status, env groups list, plugin search path (with ok/missing status), discovered plugins with tool counts
- [ ] `wtmcpctl check --workdir /custom/path` respects the workdir override
- [ ] The command is registered in `wtmcpctl --help` output
- [ ] `go test ./cmd/wtmcpctl/...` passes
- [ ] `golangci-lint run ./cmd/wtmcpctl/...` passes

## Files to create/modify

- `cmd/wtmcpctl/check.go` — new file with the check command implementation
- `cmd/wtmcpctl/main.go` — add `checkCmd` to `rootCmd.AddCommand(...)`

## Estimated diff size

~75 lines

### LINKED ISSUE #56: Add wtmcpctl check to README-wtmcpctl.md
## Description

Document the new `wtmcpctl check` command in `README-wtmcpctl.md`, following the existing format used for other commands (agent, oauth, plugin, provider, stats, vault). Place it before the `agent` section since it is a diagnostic/setup command.

**Depends on:** #55 (`wtmcpctl check` implementation)

## Acceptance Criteria

- [ ] `README-wtmcpctl.md` includes a `### check` section documenting the command
- [ ] The section follows the same format as other commands: description, usage example, expected output snippet
- [ ] The section is placed logically (before `agent` or after `version`)

## Files to modify

- `README-wtmcpctl.md` — add `check` command documentation section

## Estimated diff size

~25 lines

decko:
## Summary

Closes #55, closes #56.

Add `wtmcpctl check` command for pre-flight configuration validation.
Extract vault diagnostic logic into shared `internal/diagnostic` package,
fixing a `VaultPasswordCloser` leak in the server-side `runCheck`.

## Changes

- `cmd/wtmcpctl/check.go` — new `check` command using `plugin.Discover()`
- `internal/diagnostic/vault.go` — shared vault/credential diagnostic output (was duplicated in `cmd/wtmcp/main.go` and `cmd/wtmcpctl/check.go`)
- `internal/diagnostic/vault_test.go` — tests for password source detection, sorted output, read error handling
- `cmd/wtmcp/main.go` — replaced inline vault diagnostics with `diagnostic.Print*`, added `defer result.Close()`
- `README-wtmcpctl.md` — documented the `check` command

## Review feedback addressed

- Extracted ~160 lines of duplicated vault diagnostic code into `internal/diagnostic`
- Fixed `VaultPasswordCloser` leak in server-side `runCheck` (`defer result.Close()`)
- Sorted map iteration for deterministic output (`slices.Sort`)
- Report `f.Read` errors instead of silently skipping
- Added 9 tests in `diagnostic/vault_test.go`

### DIFF
--- a/cmd/wtmcpctl/check.go
+++ b/cmd/wtmcpctl/check.go
@@ -1,0 +1,256 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/LeGambiArt/wtmcp/internal/config"
+	"github.com/LeGambiArt/wtmcp/internal/plugin"
+	"github.com/LeGambiArt/wtmcp/internal/secrets/vault"
+)
+
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "Print diagnostic info about config and plugins",
+	Args:  cobra.NoArgs,
+	RunE:  runCtlCheck,
+}
+
+func runCtlCheck(_ *cobra.Command, _ []string) error {
+	result, err := plugin.Discover(plugin.DiscoveryOptions{
+		WorkdirOverride: globalWorkdir,
+	})
+	if err != nil {
+		return err
+	}
+	defer result.Close()
+
+	fmt.Printf("wtmcpctl %s\n", Version)
+	fmt.Printf("workdir: %s\n", result.Workdir)
+	if len(result.Config.Plugins.Enabled) > 0 {
+		fmt.Printf("plugin mode: allowlist (%d plugins)\n", len(result.Config.Plugins.Enabled))
+	} else {
+		fmt.Printf("plugin mode: default\n")
+	}
+	fmt.Printf("user plugins: %v\n", result.Config.Plugins.UserPlugins)
+
+	printCtlVaultStatus(result)
+
+	fmt.Printf("env groups: %d\n", len(result.EnvGroups))
+	for group := range result.EnvGroups {
+		fmt.Printf("  - %s\n", group)
+	}
+	if result.EnvDirError != "" {
+		fmt.Printf("env.d directory error: %s\n", result.EnvDirError)
+	}
+	if len(result.EnvErrors) > 0 {
+		fmt.Printf("env group errors: %d\n", len(result.EnvErrors))
+		for group, msg := range result.EnvErrors {
+			fmt.Printf("  - %s: %s\n", group, msg)
+		}
+	}
+
+	fmt.Printf("\nplugin search path:\n")
+	for i, dir := range result.Config.PluginDirs {
+		exists := "missing"
+		if info, statErr := os.Stat(dir); statErr == nil && info.IsDir() {
+			exists = "ok"
+		}
+		fmt.Printf("  %d. %s [%s]\n", i+1, dir, exists)
+	}
+
+	manifests := result.Manager.Manifests()
+	fmt.Printf("\ndiscovered plugins: %d\n", len(manifests))
+	var totalPrimary, totalDeferred int
+	for _, m := range manifests {
+		var primaryCount, deferredCount int
+		for _, t := range m.Tools {
+			if t.IsPrimary() {
+				primaryCount++
+			} else {
+				deferredCount++
+			}
+		}
+		totalPrimary += primaryCount
+		totalDeferred += deferredCount
+		fmt.Printf("  - %s v%s (%s)\n", m.Name, m.Version, m.Dir)
+		fmt.Printf("    handler: %s | execution: %s | tools: %d (primary: %d, deferred: %d)\n",
+			m.Handler, m.Execution, len(m.Tools), primaryCount, deferredCount)
+	}
+
+	fmt.Printf("\ntool discovery: %s\n", result.Config.Tools.Discovery)
+	fmt.Printf("primary tools: %d\n", totalPrimary)
+	fmt.Printf("deferred tools: %d\n", totalDeferred)
+
+	if len(manifests) == 0 {
+		fmt.Println("\nno plugins found. check that plugin directories contain")
+		fmt.Println("subdirectories with plugin.yaml files.")
+	}
+
+	return nil
+}
+
+// printCtlVaultStatus reports vault password configuration and per-group
+// encryption status for the check command.
+func printCtlVaultStatus(result *plugin.DiscoveryResult) {
+	cfg := result.Config
+
+	passwordSource := "not configured"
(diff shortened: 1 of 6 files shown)


cmd/wtmcpctl/check.go
```diff
@@ -0,0 +1,256 @@
...
+
+	if len(manifests) == 0 {
+		fmt.Println("\nno plugins found. check that plugin directories contain")
+		fmt.Println("subdirectories with plugin.yaml files.")
+	}
+
+	return nil
+}
+
+// printCtlVaultStatus reports vault password configuration and per-group
```
sergio-correia: `printCtlVaultStatus` and `printCtlCredentialFileStatus` are character-for-character copies of `printVaultStatus` and `printCredentialFileStatus` from `cmd/wtmcp/main.go`, differing only in the function name prefix. That is about 160 lines of security-sensitive code (vault password resolution, trial decryption, `ZeroBytes` cleanup) maintained in two places.

Divergence has already started: this version correctly adds `defer result.Close()` at line 30, but the server's `runCheck()` does not, leaking the `VaultPasswordCloser`.

Please extract these into a shared `internal/diagnostic` package (or add them as methods on `*plugin.DiscoveryResult`). Both binaries would call the shared implementation, and the server/ctl wrappers would only handle their own extra output lines (sandbox, readOnly).

cmd/wtmcpctl/check.go
```diff
@@ -0,0 +1,256 @@
...
+		fmt.Printf("plugin mode: allowlist (%d plugins)\n", len(result.Config.Plugins.Enabled))
+	} else {
+		fmt.Printf("plugin mode: default\n")
+	}
+	fmt.Printf("user plugins: %v\n", result.Config.Plugins.UserPlugins)
+
+	printCtlVaultStatus(result)
+
+	fmt.Printf("env groups: %d\n", len(result.EnvGroups))
+	for group := range result.EnvGroups {
```
sergio-correia: Go map iteration order is randomized, so running `wtmcpctl check` twice can list env groups and error entries in a different order. For a diagnostic tool where users might diff output across runs or paste it into bug reports, sorted output would be more useful. Consider collecting the map keys into a `[]string` and sorting with `slices.Sort` before iterating.

cmd/wtmcpctl/check.go
```diff
@@ -0,0 +1,256 @@
...
+			if file.IsDir() {
+				continue
+			}
+			path := filepath.Join(groupDir, file.Name())
+			f, err := os.Open(path) //nolint:gosec // credentials dir from config
+			if err != nil {
+				continue
+			}
+			header := make([]byte, 15)
+			n, _ := f.Read(header)
```
sergio-correia: `n, _ := f.Read(header)` discards the read error. If `Read` returns `n=0` due to an I/O error, the file gets silently skipped. For a diagnostic tool whose job is to surface problems, it would be better to report the error so the user knows a credential file could not be inspected. This is inherited from the server, but worth fixing here.

sergio-correia (commented):
A few things to address, mainly around extracting the duplicated vault diagnostic code into a shared package, adding docs and tests.

The `check` command is missing from `README-wtmcpctl.md`. Every other subcommand (`agent`, `oauth`, `plugin`, `provider`, `stats`, `vault`, `version`) has a documented section there. Please add a `### check` section between `### agent` and `### oauth` covering the command's purpose, the `--workdir` flag, and example output.

Other wtmcpctl commands have test files (`agent_test.go`, `credentials_google_test.go`, `stats_test.go`), but `check` has none. Please add a `check_test.go` with at least a happy-path test (temp workdir with minimal config, verify `runCtlCheck` returns no error) and a test for vault password source detection covering all three branches (file, vault IDs, env var). If the duplication is extracted first, the tests can target the shared package instead.

cmd/wtmcpctl/check.go
decko: Extracted both into `internal/diagnostic`. Server-side `Close()` leak fixed too. Both binaries now call the shared implementation.

cmd/wtmcpctl/check.go
decko: Done. `PrintEnvGroups` sorts both group names and error keys with `slices.Sort`.

cmd/wtmcpctl/check.go
decko: Fixed. Reports the error when `n=0` instead of silently skipping.

decko:
All items addressed in 589cc9d:

- Vault diagnostics extracted to `internal/diagnostic`, both binaries use shared implementation
- Server-side `Close()` leak fixed (`defer result.Close()`)
- Map iteration sorted via `slices.Sort`
- Read error reported instead of silently skipped
- Tests added in `diagnostic/vault_test.go` (9 tests: password source detection, sorted output, read errors)
- Docs tracked separately in #56

decko:
This PR now also includes the `wtmcpctl check` documentation in README-wtmcpctl.md, closing #56.

sergio-correia (approved):
LGTM, thanks!

PR #75, opened 2026-05-26
fix: include full commit message in gitlab_get_commits

sergio-correia:
The tool only returned the commit title (first line), making it impossible to access information in the commit body such as ticket references, changelog entries, or sign-off lines. The go-gitlab API already provides the full message — pass it through to match the behavior of github_get_pr_commits.


rjeffman (approved):
LGTM.

PR #87, opened 2026-06-03
feat(wtmcpctl): add user-plugins command to manage user plugin support

rjeffman:
Add wtmcpctl plugin user-plugins subcommand with enable, disable, and status actions to manage the plugins.user_plugins configuration option.

This provides a convenient CLI interface for toggling user plugin support instead of manually editing config.yaml.

Commands:
- wtmcpctl plugin user-plugins enable: set user_plugins to true
- wtmcpctl plugin user-plugins disable: set user_plugins to false
- wtmcpctl plugin user-plugins status: show current state and list user plugins

Implementation:
- Added updateConfigBool() helper in common.go for atomic boolean config updates
- Added user-plugins command hierarchy in plugins.go with three subcommands
- Status command shows enabled/disabled state and lists discovered user plugins

### DIFF
--- a/cmd/wtmcpctl/plugins.go
+++ b/cmd/wtmcpctl/plugins.go
@@ -74,13 +74,50 @@ except those in plugins.disabled).`,
 	RunE:              runPluginsOnly,
 }
 
+var userPluginsCmd = &cobra.Command{
+	Use:   "user-plugins",
+	Short: "Manage user plugin support",
+	Long: `Enable or disable loading of user plugins from {workdir}/plugins/.
+
+User plugins are disabled by default for security. When enabled, plugins
+in your workdir/plugins/ directory will be loaded alongside system plugins.
+
+User plugins cannot:
+- Override system plugins
+- Declare authentication providers
+- Claim credential groups owned by system plugins
+- Use symlinks or hardlinks in handler paths`,
+}
+
+var userPluginsEnableCmd = &cobra.Command{
+	Use:   "enable",
+	Short: "Enable user plugin support",
+	Args:  cobra.NoArgs,
+	RunE:  runUserPluginsEnable,
+}
+
+var userPluginsDisableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable user plugin support",
+	Args:  cobra.NoArgs,
+	RunE:  runUserPluginsDisable,
+}
+
+var userPluginsStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Show user plugin support status",
+	Args:  cobra.NoArgs,
+	RunE:  runUserPluginsStatus,
+}
+
 func init() {
 	pluginsListCmd.Flags().BoolP("plain", "p", false,
 		"Plain text output (no colors or borders)")
 	pluginsOnlyCmd.Flags().Bool("clear", false,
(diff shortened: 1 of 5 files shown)


cmd/wtmcpctl/plugins.go
```diff
@@ -645,3 +682,81 @@ func completeAllPlugins(
...
+		return err
+	}
+
+	fmt.Printf("User plugin support disabled in %s\n", result.ConfigPath)
+	fmt.Println("Restart wtmcp for changes to take effect.")
+	return nil
+}
+
+func runUserPluginsStatus(_ *cobra.Command, _ []string) error {
+	result, err := getDiscoveryResult()
```
sergio-correia: This should use `getPluginsDiscoveryResult()` instead of `getDiscoveryResult()`. The filtered variant moves individually disabled plugins into `configDisabled`, so they won't show up in `Manifests()`. If a user disables one of their user plugins via `plugin disable`, this command will undercount.

Every other plugin-listing command (`list`, `enable`, `disable`, `only`) uses `getPluginsDiscoveryResult()` for this reason.

```suggestion
	result, err := getPluginsDiscoveryResult()
```

cmd/wtmcpctl/plugins.go
```diff
@@ -645,3 +682,81 @@ func completeAllPlugins(
...
+	}
+
+	fmt.Printf("User plugin support: %s\n", statusStyle.Render(status))
+	fmt.Printf("Config file: %s\n", result.ConfigPath)
+	fmt.Printf("User plugin directory: %s/plugins/\n", result.Workdir)
+
+	if result.Config.Plugins.UserPlugins {
+		// Count user plugins if enabled
+		userPluginCount := 0
+		for _, m := range result.Manager.Manifests() {
```
sergio-correia, line 743: `Manifests()` is called twice here (once to count, once to print). Each call clones every manifest, and since Go map iteration is non-deterministic, the listed plugins will appear in random order on each run. The other commands avoid this via `buildPluginEntries` which stores the result and sorts it.

Would be cleaner to call `Manifests()` once, collect the user plugins into a slice, and sort by name before printing.

cmd/wtmcpctl/plugins.go
```diff
@@ -645,3 +682,81 @@ func completeAllPlugins(
...
+	return nil
+}
+
+func runUserPluginsDisable(_ *cobra.Command, _ []string) error {
+	result, err := getDiscoveryResult()
+	if err != nil {
+		return err
+	}
+
+	disabled := false
```
sergio-correia: The name `disabled` holding the value `false` is easy to misread. A maintainer who "fixes" this to `disabled := true` would silently enable user plugins on the disable path. Maybe rename to `val := false`.

cmd/wtmcpctl/plugins.go
rjeffman: fixed.

cmd/wtmcpctl/plugins.go
rjeffman: fixed.

cmd/wtmcpctl/plugins.go
rjeffman: fixed.

sergio-correia (approved):
LGTM, thanks!

PR #93, opened 2026-06-07
sandbox: make arapuca sandbox mandatory and self-contained

sergio-correia:
- Make the arapuca Landlock+seccomp sandbox built and enabled by default, removing the runtime config toggle
- Add selfexec mode so wtmcp acts as its own sandbox trampoline — no external arapuca binary needed
-  Auto-detect system libarapuca via pkg-config, falling back to building from the third_party/arapuca submodule
- Route pre-commit hooks through make targets to handle PKG_CONFIG_PATH automatically
- Update CI to build libarapuca and run both sandboxed and nosandbox stub


rjeffman (commented):
LGTM

rjeffman (approved):

PR #94, opened 2026-06-08
security: harden input sanitization, credential paths, and plugin validation

sergio-correia:
- Content sanitization pipeline for tool output (UTF-8 cleanup, Unicode format char stripping, HTML comment removal) with nonce-based output tagging and audience annotations
- Credential path containment via EvalSymlinks + prefix checks on OAuth2 and Google token paths; symlink/permission validation on load and save
- Three-layer plugin handler validation (manifest, discovery, launch-time) with symlink, hardlink, and intermediate directory checks for user plugins
- dry_run defaults to true on all remaining write tools; write-tool exclusion from tool_search; max output size cap with UTF-8-safe truncation
- Domain validation with IDNA normalization for dynamic auth bindings; deterministic domain capping
- User plugin trust prefix on both enabled and disabled tool descriptions 
- Fuzz tests for sanitization, credential scrubbing, path confinement, domain validation, output framing, SSRF IP validation, and plugin name validation
- Documentation for macOS TOCTOU limitations, GitHub PAT scopes, and Google OAuth read-only alternatives

### DIFF
--- a/internal/auth/oauth2.go
+++ b/internal/auth/oauth2.go
@@ -314,15 +314,25 @@ func resolveCredentialPath(path, credentialsDir string) (string, error) {
 		}
 		base = filepath.Join(home, ".config", "wtmcp", "credentials")
 	}
+	if resolvedBase, err := filepath.EvalSymlinks(base); err == nil {
+		base = resolvedBase
+	}
 	var resolved string
 	if filepath.IsAbs(path) {
 		resolved = filepath.Clean(path)
 	} else {
 		resolved = filepath.Clean(filepath.Join(base, path))
 	}
+	if r, err := filepath.EvalSymlinks(resolved); err == nil {
+		resolved = r
+	} else if dir := filepath.Dir(resolved); dir != resolved {
+		if rd, err := filepath.EvalSymlinks(dir); err == nil {
+			resolved = filepath.Join(rd, filepath.Base(resolved))
+		}
+	}
 	cleanBase := filepath.Clean(base)
 	if resolved != cleanBase &&
-		!strings.HasPrefix(resolved, cleanBase+string(os.PathSeparator)) {
+		!strings.HasPrefix(resolved, cleanBase+string(filepath.Separator)) {
 		return "", fmt.Errorf("credential path escapes credentials directory: %s", path)
 	}
 	return resolved, nil
--- a/internal/auth/oauth2.go
+++ b/internal/auth/oauth2.go
@@ -192,6 +192,12 @@ func loadToken(path string) (*oauth2.Token, error) {
 }
 
 func saveToken(path string, tok *oauth2.Token) error {
+	if _, err := os.Lstat(path); err == nil {
+		if err := config.RejectSymlink(path); err != nil {
+			return fmt.Errorf("token save: %w", err)
+		}
+	}
+
 	tj := tokenJSON{
 		AccessToken:  tok.AccessToken,
 		TokenType:    tok.TokenType,
@@ -251,7 +257,17 @@ type credentialsData struct {
 }
 
 func loadOAuth2Config(path string, scopes []string) (*oauth2.Config, error) {
-	data, err := os.ReadFile(path) //nolint:gosec // credential file path from config
+	if err := config.RejectSymlink(path); err != nil {
+		return nil, fmt.Errorf("credentials file: %w", err)
+	}
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if err := config.CheckPermissions(path, info); err != nil {
+		return nil, fmt.Errorf("credentials file: %w", err)
+	}
+	data, err := os.ReadFile(path) //nolint:gosec // validated above
 	if err != nil {
 		return nil, err
 	}
--- a/internal/auth/oauth2.go
+++ b/internal/auth/oauth2.go
@@ -316,6 +316,8 @@ func resolveCredentialPath(path, credentialsDir string) (string, error) {
 	}
 	if resolvedBase, err := filepath.EvalSymlinks(base); err == nil {
 		base = resolvedBase
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("resolve credentials base: %w", err)
 	}
 	var resolved string
 	if filepath.IsAbs(path) {
@@ -325,9 +327,13 @@ func resolveCredentialPath(path, credentialsDir string) (string, error) {
 	}
 	if r, err := filepath.EvalSymlinks(resolved); err == nil {
 		resolved = r
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("resolve credential path: %w", err)
 	} else if dir := filepath.Dir(resolved); dir != resolved {
 		if rd, err := filepath.EvalSymlinks(dir); err == nil {
 			resolved = filepath.Join(rd, filepath.Base(resolved))
+		} else if !os.IsNotExist(err) {
+			return "", fmt.Errorf("resolve credential parent dir: %w", err)
 		}
 	}
 	cleanBase := filepath.Clean(base)
--- a/internal/domaincheck/validate.go
+++ b/internal/domaincheck/validate.go
@@ -24,7 +24,7 @@ import (
 func Normalize(domain string) (string, error) {
 	domain = strings.TrimSpace(domain)
 	domain = strings.ToLower(domain)
-	domain = strings.TrimSuffix(domain, ".")
+	domain = strings.TrimRight(domain, ".")
 
 	// Convert Unicode domains to ASCII using IDNA (Internationalized
 	// Domain Names in Applications). This handles punycode (xn--) and
--- a/internal/google/auth.go
+++ b/internal/google/auth.go
@@ -5,11 +5,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
+	"sync"
 	"time"
 
+	"github.com/LeGambiArt/wtmcp/internal/config"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 )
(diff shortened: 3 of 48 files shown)


internal/domaincheck/validate.go
```diff
@@ -24,7 +24,7 @@ import (
 func Normalize(domain string) (string, error) {
 	domain = strings.TrimSpace(domain)
 	domain = strings.ToLower(domain)
-	domain = strings.TrimSuffix(domain, ".")
+	domain = strings.TrimRight(domain, ".")
```
rjeffman, line 27: Don't we want to have the dot at the end no mark the domain as "full name"?

sergio-correia: Good question! The change from `TrimSuffix` to `TrimRight` was actually a bug fix caught by fuzz testing. `TrimSuffix(domain, ".")` only strips a single trailing dot, so an input like `"0000.."` would normalize to `"0000."`, still carrying a dot. That mismatch could let a crafted domain slip past allowlist comparisons.

`TrimRight` strips all trailing dots, so we get a consistent canonical form regardless of how many dots are appended. Since `Normalize` is only used for internal comparison (not for constructing DNS queries), we don't need to preserve the FQDN dot. We just need example.com. and example.com to resolve to the same string.

internal/google/auth.go
```diff
@@ -141,10 +196,37 @@ func saveToken(path string, tok *oauth2.Token) error {
...
+	if err := tmp.Sync(); err != nil {
+		tmp.Close()        //nolint:errcheck,gosec // best-effort cleanup
+		os.Remove(tmpName) //nolint:errcheck,gosec // best-effort cleanup
+		return err
+	}
+	if err := tmp.Close(); err != nil {
+		os.Remove(tmpName) //nolint:errcheck,gosec // best-effort cleanup
+		return err
+	}
+	return os.Rename(tmpName, path)
```
decko, line 228: Credential leak: if `Rename` fails (cross-device, permissions), `.token-*.tmp` with the full OAuth token persists on disk. Every other error path cleans up.

```go
if err := os.Rename(tmpName, path); err != nil {
    os.Remove(tmpName)
    return fmt.Errorf("rename token file: %w", err)
}
return nil
```

Same applies to `internal/auth/oauth2.go` `saveToken`.

internal/auth/oauth2.go
```diff
@@ -298,15 +314,25 @@ func resolveCredentialPath(path, credentialsDir string) (string, error) {
 		}
 		base = filepath.Join(home, ".config", "wtmcp", "credentials")
 	}
+	if resolvedBase, err := filepath.EvalSymlinks(base); err == nil {
```
decko, line 317: All three `EvalSymlinks` calls silently swallow non-`ErrNotExist` errors (permission denied, circular symlinks, I/O). When they fail, prefix check runs on unresolved paths while reads follow symlinks — potential containment bypass.

```go
if resolvedBase, err := filepath.EvalSymlinks(base); err == nil {
    base = resolvedBase
} else if !os.IsNotExist(err) {
    return "", fmt.Errorf("resolve credentials base: %w", err)
}
```

plugins/google-drive/tools.go
```diff
@@ -513,24 +513,38 @@ func buildSearchQuery(text string, inNameOnly bool, mimeTypes, owners []string,
...
+
+	if !filepath.IsAbs(filePath) {
+		filePath = filepath.Join(sessionDir, filePath)
+	}
+
+	resolvedSession, err := filepath.EvalSymlinks(sessionDir)
+	if err != nil {
+		return "", fmt.Errorf("resolve session dir: %w", err)
+	}
+	resolved, err := filepath.EvalSymlinks(filePath)
```
decko, line 529: `EvalSymlinks` fails for nonexistent files. Currently only used by `toolUploadFile` (existing files), but the function name suggests general-purpose confinement. Export tools creating new output files would break.

Add parent-dir fallback (same pattern as `resolveCredentialPath`):
```go
if err != nil {
    dir := filepath.Dir(filePath)
    if rd, err2 := filepath.EvalSymlinks(dir); err2 == nil {
        resolved = filepath.Join(rd, filepath.Base(filePath))
    } else {
        return "", fmt.Errorf("resolve path: %w", err)
    }
}
```

internal/server/server.go
```diff
@@ -420,15 +434,19 @@ func registerDisabledPluginTools(srv *mcpserver.MCPServer, disabled map[string]p
...
+				prefix, sanitizeContent(dp.Reason), pluginName, sanitizeContent(toolDef.Description),
 			)
 
 			reason := dp.Reason
 			name := pluginName
 			srv.AddTool(tool, func(_ context.Context, _ mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-				return mcp.NewToolResultError(fmt.Sprintf(
+				return frameErrorResult(fmt.Sprintf(
 					"[DISABLED] %s\n\nAfter fixing, run: plugin_reload(name=\"%s\")",
 					reason, name,
```
decko, line 442: `reason` goes through `sanitizeContent()` in the description above but not `auditor.ScrubErrorText()` here. Disable reasons from plugin errors can contain credential fragments. At minimum: `sanitizeContent(reason)` for consistency.

internal/plugin/manager.go
```diff
@@ -1003,8 +1018,8 @@ func (m *Manager) sanitizeReason(reason string) string {
 	if m.workdir != "" {
 		reason = strings.ReplaceAll(reason, m.workdir+"/", "")
 	}
-	if m.envDir != "" && !strings.HasPrefix(m.envDir, m.workdir+"/") {
-		reason = strings.ReplaceAll(reason, m.envDir, "env.d")
+	if m.envDir != "" && (m.workdir == "" || !strings.HasPrefix(m.envDir, m.workdir+"/")) {
+		reason = strings.ReplaceAll(reason, m.envDir+"/", "env.d/")
```
decko, line 1022: Appending `/` prevents partial component matching (good), but bare `envDir` without trailing slash is no longer sanitized. E.g. `"/opt/secrets/env.d has mode 0755"` leaks the path.

Do both (longer form first):
```go
reason = strings.ReplaceAll(reason, m.envDir+"/", "env.d/")
reason = strings.ReplaceAll(reason, m.envDir, "env.d")
```

internal/plugin/manager.go
```diff
@@ -1159,19 +1174,54 @@ func discoverSSHPublicKeys() string {
...
+		if !hasSSHKeyPrefix(k) {
+			continue
+		}
+		keys = append(keys, k)
+		if len(keys) >= maxSSHKeys {
+			log.Printf("WARNING: SSH key discovery capped at %d keys", maxSSHKeys)
+			break
 		}
 	}
 	return strings.Join(keys, "\n")
```
decko, line 1215: If all `.pub` files fail validation (symlinks, oversized, wrong prefix), returns empty with no log. User's Testing Farm reservation gets no SSH keys with no indication why.

```go
if len(matches) > 0 && len(keys) == 0 {
    log.Printf("WARNING: found %d SSH .pub files but none passed validation", len(matches))
}
```

internal/google/auth.go
```diff
@@ -82,29 +114,44 @@ func NewHTTPClient(ctx context.Context, tokenFile string, scopes []string) (*htt
...
 	if err != nil {
 		return nil, err
 	}
 
-	// Save if the token changed (was refreshed)
 	if s.lastToken == nil || tok.AccessToken != s.lastToken.AccessToken {
 		s.lastToken = tok
-		_ = saveToken(s.tokenPath, tok) // best-effort save
+		if err := saveToken(s.tokenPath, tok); err != nil {
+			log.Printf("google: failed to persist refreshed token: %v", err)
```
decko, line 135: Security-relevant `saveToken` failures (symlink rejection = possible attack) logged at the same level as normal failures (disk full). Consider elevating symlink rejections — they may indicate an active attack on the credential path.

decko (commented):
Thorough hardening PR — well-structured layered defenses and good fuzz coverage. Seven inline comments (mostly gap-fills and a one-liner credential leak fix), plus two design questions:

1. **Is `auditor` truly optional?** The nil-checks trade a panic (fail-closed) for unscrubbed credential leakage to the LLM (fail-open). Either make it a required non-nil dependency or add a fallback scrubber.

2. **Breaking changes need migration notes:** `toolNamePattern` rejects hyphens in tool names, `ValidateUserHandler` rejects nonexistent handlers. Both could break third-party/user plugins on upgrade.

internal/google/auth.go
sergio-correia: Fixed. Added `defer os.Remove(tmpName)` after creating the temp file, matching the pattern already used in `oauth2.go`. The defer covers all error paths including the rename failure. Note that `oauth2.go` already had this, so the fix was only needed here.

internal/auth/oauth2.go
sergio-correia: Fixed. All three `EvalSymlinks` calls now propagate non-`ErrNotExist` errors instead of silently falling through to the prefix check. Agreed this was the most important one.

plugins/google-drive/tools.go
sergio-correia: Leaving this as-is for now. The only caller is `toolUploadFile`, which always operates on existing files. Adding the parent-dir fallback would be complexity for a use case that doesn't exist today. If we add export tools that create new output files, we can revisit.

internal/server/server.go
sergio-correia: Fixed. `registerDisabledPluginTools` now receives auditor and scrubs the reason via `ScrubErrorText` before returning it in the handler. Also plumbed auditor through `SwapStartFailedTools` for the same path.

internal/plugin/manager.go
sergio-correia: Good catch. Fixed by doing both replacements (longer form first):
```go
reason = strings.ReplaceAll(reason, m.envDir+"/", "env.d/")
reason = strings.ReplaceAll(reason, m.envDir, "env.d")
```
Same treatment for `credentialsDir`.

internal/plugin/manager.go
sergio-correia: Fixed. Added a warning when matches were found but all keys failed validation:
```go
if len(matches) > 0 && len(keys) == 0 {
    log.Printf("WARNING: found %d SSH .pub files but none passed validation", len(matches))
}
```

internal/google/auth.go
sergio-correia: The codebase uses Go's standard `log` package, which doesn't have level support. Elevating symlink rejections would mean pulling in a structured logging dependency, which feels out of scope for this PR. Something to consider if we ever migrate to `slog` for the main server logs.

sergio-correia:
> Thorough hardening PR — well-structured layered defenses and good fuzz coverage. Seven inline comments (mostly gap-fills and a one-liner credential leak fix), plus two design questions:
> 
>     1. **Is `auditor` truly optional?** The nil-checks trade a panic (fail-closed) for unscrubbed credential leakage to the LLM (fail-open). Either make it a required non-nil dependency or add a fallback scrubber.
> 
>     2. **Breaking changes need migration notes:** `toolNamePattern` rejects hyphens in tool names, `ValidateUserHandler` rejects nonexistent handlers. Both could break third-party/user plugins on upgrade.

Thanks for the review!

On auditor being optional: fixed. All `Logger` methods now have nil-receiver guards (same pattern `ScrubErrorText` already had), and the `if auditor != nil` branches in server.go are gone. Also caught an inverted condition in `Close()` that would have panicked on a nil receiver, so thanks for flagging this.

On breaking changes: `toolNamePattern` rejecting hyphens and `ValidateUserHandler` rejecting nonexistent handlers are both intentional tightening. All existing plugins use underscores already. Happy to add a migration note in the release docs if you think third-party plugin authors need the heads-up.

decko (approved):
LTTM 🚀

PR #95, opened 2026-06-08
feat(jira): add set_issue_type and get_issue_types tools

decko:
## Summary
- Add `jira_set_issue_type` write tool to change an issue's type (e.g. Story → Bug, Task → Epic) via `PUT /rest/api/2/issue/{key}` with `dry_run=true` default
- Add `jira_get_issue_types` read tool to list available issue types for a project (or all instance types), cached for 1 hour — intended as a discovery step before `set_issue_type`
- 12 new tests (5 read, 7 write) covering dry_run, success, HTTP errors, cache behavior, and input validation

## Test plan
- [x] `pytest plugins/jira/tests/ -v` — 468 passed
- [x] `ruff check` — clean
- [ ] Manual: call `jira_get_issue_types` with a project key, verify returned types
- [ ] Manual: call `jira_set_issue_type` with `dry_run=true`, verify preview
- [ ] Manual: call `jira_set_issue_type` with `dry_run=false`, verify type changes in Jira

### DIFF
--- a/plugins/jira/tools_read.py
+++ b/plugins/jira/tools_read.py
@@ -297,8 +297,7 @@ def get_issue_types(params):
 
     result = {
         "issue_types": [
-            {"id": t.get("id"), "name": t.get("name"), "subtask": t.get("subtask", False)}
-            for t in raw_types
+            {"id": t.get("id"), "name": t.get("name"), "subtask": t.get("subtask", False)} for t in raw_types
         ],
     }
     if project_key:
--- a/plugins/jira/tools_read.py
+++ b/plugins/jira/tools_read.py
@@ -10,6 +10,7 @@ from helpers import (
     http_error,
     is_user_alias,
     validate_issue_key,
+    validate_project_key,
 )
 
 
@@ -277,9 +278,10 @@ def get_link_types(_params):
 
 def get_issue_types(params):
     """Get available issue types for a project, cached for 1 hour."""
-    project_key = params.get("project_key", "")
+    raw_key = params.get("project_key", "")
+    project_key = validate_project_key(raw_key) if raw_key else ""
 
-    cache_key = f"issue_types:{project_key}" if project_key else "issue_types:all"
+    cache_key = f"issue_types:project:{project_key}" if project_key else "issue_types:global"
     cached = handler.cache_get(cache_key)
     if cached:
         return cached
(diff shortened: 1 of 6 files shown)


sergio-correia (commented):
A couple of things to fix before merging. Please also run the pre-commit hook to fix the ruff issue.

plugins/jira/tools_read.py
```diff
@@ -275,6 +275,38 @@ def get_link_types(_params):
     return body
 
 
+def get_issue_types(params):
+    """Get available issue types for a project, cached for 1 hour."""
+    project_key = params.get("project_key", "")
+
+    cache_key = f"issue_types:{project_key}" if project_key else "issue_types:all"
```
sergio-correia: The cache key `"issue_types:all"` collides with a project literally named `ALL` (or lowercase `all`, since `project_key` isn't uppercased here). That project would get the instance-wide cache entry, or worse, a project-scoped entry would be served for the global query.

```suggestion
    cache_key = f"issue_types:project:{project_key}" if project_key else "issue_types:global"
```

plugins/jira/tools_read.py
```diff
@@ -275,6 +275,38 @@ def get_link_types(_params):
...
+    cached = handler.cache_get(cache_key)
+    if cached:
+        return cached
+
+    if project_key:
+        status, body, _ = handler.http("GET", f"/rest/api/2/project/{project_key}")
```
sergio-correia, line 290: `project_key` is interpolated into the URL path without validation. Unlike `issue_key` (which goes through `validate_issue_key` with a strict regex), this accepts arbitrary strings including `../` sequences. A value like `X/../../rest/api/2/myself` would hit an unintended endpoint after path normalization.

The same gap exists for `board_id` in `tools_sprint.py`, but worth fixing here to establish the right pattern. A `validate_project_key` helper in `helpers.py` with `^[A-Z][A-Z0-9_]+$` would match Jira's project key format.

plugins/jira/tools_read.py
decko: Fixed in 7c69fb1 — cache keys now use `issue_types:project:{key}` / `issue_types:global`, and `project_key` is validated + uppercased before use.

plugins/jira/tools_read.py
decko: Added `validate_project_key()` in helpers.py with `^[A-Z][A-Z0-9_]+$` regex — same commit (7c69fb1). Agreed on the `board_id` gap in `tools_sprint.py`, will address separately.

sergio-correia (approved):

PR #106, opened 2026-06-11
sandbox: allow EvalSymlinks to resolve paths within session directory

sergio-correia:
Fix `file_path` parameter failing with `lstat /Users: operation not permitted` when `wtmcp`runs sandboxed: the confinement check's `EvalSymlinks` call walks from root, hitting path components not individually listed in the sandbox profile.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"maps"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -94,12 +95,24 @@ func New(version string, manager *plugin.Manager, cfg *config.Config, index *Too
 	if cfg.Security.ElicitationEnabled() {
 		opts = append(opts, mcpserver.WithElicitation())
 	}
+	var instructions []string
 	if !sandboxBuilt {
-		opts = append(opts, mcpserver.WithInstructions(
+		instructions = append(instructions,
 			"WARNING: This server is running WITHOUT sandbox isolation. "+
 				"Plugin processes have unrestricted access to the filesystem and network. "+
 				"This configuration is unsafe for production use. "+
-				"Exercise caution with tools that write files or make network requests."))
+				"Exercise caution with tools that write files or make network requests.")
+	}
+	if sd := manager.SessionDir(); sd != "" {
+		instructions = append(instructions,
+			fmt.Sprintf("Session directory: %s — "+
+				"file_path parameters in tools that accept local files "+
+				"resolve relative paths against this directory. "+
+				"Output files are written to %s/.wtmcp-data/<plugin>/.",
+				sd, sd))
+	}
+	if len(instructions) > 0 {
+		opts = append(opts, mcpserver.WithInstructions(strings.Join(instructions, "\n\n")))
 	}
 	srv := mcpserver.NewMCPServer("wtmcp", version, opts...)
 
@@ -138,6 +151,9 @@ func New(version string, manager *plugin.Manager, cfg *config.Config, index *Too
 
 	// Register context files as MCP resources
 	registerContextResources(srv, manager, collector)
+
+	// Register session directory info as a server-level resource
+	registerSessionResource(srv, manager, collector)
 
 	// Register resources from resource provider plugins
 	RegisterPluginResources(srv, manager, collector)
@@ -814,6 +830,43 @@ func registerPluginContextResources(srv *mcpserver.MCPServer, manifest *plugin.M
 			},
 		)
 	}
+}
+
+func registerSessionResource(srv *mcpserver.MCPServer, mgr *plugin.Manager, collector *stats.Collector) {
+	sd := mgr.SessionDir()
+	if sd == "" {
+		return
+	}
+	const uri = "wtmcp://server/session"
+	srv.AddResource(
+		mcp.NewResource(uri, "Session directory info",
+			mcp.WithResourceDescription("Session directory, output paths, and file_path resolution"),
+			mcp.WithMIMEType("text/plain"),
+		),
+		func(_ context.Context, _ mcp.ReadResourceRequest) ([]mcp.ResourceContents, error) {
+			var b strings.Builder
+			fmt.Fprintf(&b, "Session directory: %s\n\n", sd)
+			fmt.Fprintf(&b, "Tools that accept a file_path parameter resolve relative paths against this directory.\n")
+			fmt.Fprintf(&b, "Output files are written under: %s/.wtmcp-data/<plugin-name>/\n\n", sd)
+			fmt.Fprintf(&b, "Per-plugin output directories:\n")
+			names := mgr.LoadedPlugins()
+			sort.Strings(names)
+			for _, name := range names {
+				fmt.Fprintf(&b, "  %s: %s/.wtmcp-data/%s/\n", name, sd, name)
+			}
+			content := b.String()
+			if collector != nil {
+				collector.RecordResourceRead(uri, "server", "session", content)
+			}
+			return []mcp.ResourceContents{
+				mcp.TextResourceContents{
+					URI:      uri,
+					MIMEType: "text/plain",
+					Text:     content,
+				},
+			}, nil
+		},
+	)
 }
 
 // processToolActions handles side effects declared in tool results.
--- a/plugins/bugzilla/helpers.go
+++ b/plugins/bugzilla/helpers.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/LeGambiArt/wtmcp/pkg/handler"
+	"github.com/LeGambiArt/wtmcp/pkg/pathutil"
 )
 
 var cfg struct {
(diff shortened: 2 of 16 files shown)


plugins/bugzilla/helpers.go
```diff
@@ -230,17 +238,28 @@ func confineRead(filePath string, allowedDirs ...string) (string, error) {
...
+func evalOrClean(path string) string {
+	if resolved, err := filepath.EvalSymlinks(path); err == nil {
+		return resolved
+	}
+	abs, err := filepath.Abs(filepath.Clean(path))
+	if err != nil {
+		return filepath.Clean(path)
+	}
+	return abs
+}
```
decko: **CRITICAL: Silent error swallowing in security-critical path**

Discards ALL `EvalSymlinks` errors — not just sandbox `EPERM`, but also `ELOOP`, `ENOENT`, `EIO`, etc. No logging, no way for callers to distinguish "fully resolved" from "best-effort."

Outside the sandbox, a non-existent `allowedDir` silently resolves to a cleaned path that may not represent a real directory.

Suggestions:
- Log when `EvalSymlinks` fails (audit trail for incident analysis)
- Return `(string, bool)` so callers know resolution level
- Extract to shared internal package — identical copy exists in `google-docs/tools.go` and `google-drive/tools.go`

plugins/bugzilla/helpers.go
```diff
@@ -211,14 +211,22 @@ func confineRead(filePath string, allowedDirs ...string) (string, error) {
...
+	// OS sandbox enforces symlink confinement at the kernel level.
+	// When it succeeds (unsandboxed), it provides defense-in-depth.
+	absFile, err := filepath.Abs(filepath.Clean(filePath))
 	if err != nil {
-		return "", fmt.Errorf("resolve path: %w", err)
+		return "", fmt.Errorf("abs path: %w", err)
+	}
+	if resolved, err := filepath.EvalSymlinks(absFile); err == nil {
+		absFile = resolved
 	}
```
decko: **HIGH: Symlink bypass when unsandboxed**

If `EvalSymlinks` fails for a non-sandbox reason (e.g. `ELOOP`), `absFile` keeps the unresolved symlink path. Then `os.Stat` (line 229) follows the symlink and succeeds, but `HasPrefix` (line 242) checks the unresolved string — a symlink under the allowed dir pointing outside passes.

Old code hard-failed on `EvalSymlinks` error. New code silently falls through.

Suggestion: when unsandboxed, treat `EvalSymlinks` failure as a hard error. Only fall back for sandbox-specific denial.

plugins/bugzilla/helpers.go
```diff
@@ -230,17 +238,28 @@ func confineRead(filePath string, allowedDirs ...string) (string, error) {
 		if dir == "" {
 			continue
 		}
-		resolvedDir, err := filepath.EvalSymlinks(dir)
-		if err != nil {
-			continue
-		}
-		if strings.HasPrefix(resolved, resolvedDir+string(os.PathSeparator)) {
-			return resolved, nil
+		absDir := evalOrClean(dir)
```
decko: **Asymmetric resolution**

File path (lines 221-227) uses inline `Abs(Clean())` + conditional `EvalSymlinks`. Allowed dir uses `evalOrClean()`. Different code paths can produce different resolution levels.

Example: `EvalSymlinks` succeeds for file (`/tmp/foo` → `/private/tmp/foo`) but fails for dir inside `evalOrClean` → `absDir = "/tmp"`. Prefix check fails on a valid path with a confusing error.

Suggestion: use the same resolution for both sides. Include both resolved values in the error message.

plugins/bugzilla/helpers_test.go
```diff
@@ -357,12 +357,11 @@ func TestConfineRead(t *testing.T) {
...
 		}
-		resolvedDir, err := filepath.EvalSymlinks(dir)
-		if err != nil {
-			t.Fatal(err)
-		}
-		if !strings.HasPrefix(got, resolvedDir) {
-			t.Errorf("result %q not under dir %q", got, resolvedDir)
+		// EvalSymlinks may resolve the path (e.g. /tmp -> /private/tmp
+		// on macOS), so use evalOrClean for the expected prefix.
+		expected := evalOrClean(dir)
```
decko: **Test validates against itself**

Uses `evalOrClean(dir)` to compute the expected value — same function the production code uses. A bug in `evalOrClean` produces the same wrong result in both, test passes.

Use `filepath.EvalSymlinks` directly here (works fine outside sandbox in tests).

internal/server/server.go
```diff
@@ -816,6 +832,43 @@ func registerPluginContextResources(srv *mcpserver.MCPServer, manifest *plugin.M
...
+			return []mcp.ResourceContents{
+				mcp.TextResourceContents{
+					URI:      uri,
+					MIMEType: "text/plain",
+					Text:     content,
+				},
+			}, nil
+		},
+	)
+}
```
decko, line 870: **No test coverage**

36 new lines, zero tests. `TestNewSandboxInactiveInjectsWarning` in `discovery_test.go` provides a template.

Worth covering: registered when `SessionDir()` non-empty, skipped when empty, content includes per-plugin dirs, names sorted.

decko (commented):
Specialist review (code-reviewer, silent-failure-hunter, pr-test-analyzer).

Core security model is sound. Concerns around silent error swallowing and asymmetric path resolution when running unsandboxed.

plugins/google-drive/tools.go
```diff
@@ -522,29 +522,40 @@ func confineToSessionDir(filePath string) (string, error) {
...
+	}
+	abs, err := filepath.Abs(filepath.Clean(path))
+	if err != nil {
+		return filepath.Clean(path)
 	}
-	return absResolved, nil
+	return abs
 }
 
 var errNoWriteScope = fmt.Errorf("write operations require re-authorization with drive (read-write) scope; " +
```
decko: Third copy of `evalOrClean`. Same silent-swallowing and asymmetric-resolution concerns — see bugzilla comment. Worth extracting to a shared `internal/` package for coordinated fixes.

decko:
## Review Summary

**Architecture**: The `Clean+Abs` primary path with best-effort `EvalSymlinks` is the right approach for sandbox compatibility.

**Key concern**: All `EvalSymlinks` failures are treated identically, but "sandbox denied lstat on ancestor" (safe to fall back) and "filesystem error / symlink loop / non-existent path" (not safe) have different security implications. Discriminating between these when running unsandboxed would address most findings.

**Findings by severity**:

| Severity | Count | Theme |
|----------|-------|-------|
| Critical | 1 | `evalOrClean` silently swallows all errors, no logging |
| High | 1 | Symlink bypass possible when unsandboxed + non-sandbox `EvalSymlinks` failure |
| Medium | 3 | Asymmetric resolution paths, self-validating test, no `registerSessionResource` tests |
| Info | 1 | `evalOrClean` duplicated 3x, worth extracting |

See inline comments for details.

sergio-correia:
@decko: issues should have been address, please take another look.

decko:
All six first-round findings are properly addressed. `ResolvePath` is well-designed — EPERM-only fallback is the right call.

One new issue: the call sites re-introduce silent error suppression via `_ =`, which can bypass confinement. See inline comment.

plugins/google-drive/tools.go
```diff
@@ -522,29 +523,20 @@ func confineToSessionDir(filePath string) (string, error) {
...
-	if err != nil {
-		return "", fmt.Errorf("resolve path: %w", err)
-	}
-
-	absSession, err := filepath.Abs(resolvedSession)
-	if err != nil {
-		return "", fmt.Errorf("abs session dir: %w", err)
+	absSession := resolvedSessionDir
+	if absSession == "" {
+		absSession, _ = pathutil.ResolvePath(sessionDir)
```
decko: **Confinement bypass when sessionDir is set but unresolvable**

Same pattern in `google-docs/tools.go` `readFileForWrite()`.

If `resolvedSessionDir` is empty (init failed) and this per-call `ResolvePath` also fails, `absSession` stays `""`. The prefix check becomes `strings.HasPrefix(absFile, "/")` — true for every absolute path. Confinement is fully bypassed.

The original bug was `evalOrClean` silently swallowing errors. `ResolvePath` fixed that, but these call sites re-introduce the same class of bug by discarding its error return.

Fix — check the error on the fallback:
```go
absSession := resolvedSessionDir
if absSession == "" {
    var err error
    absSession, err = pathutil.ResolvePath(sessionDir)
    if err != nil {
        return "", fmt.Errorf("resolve session dir: %w", err)
    }
}
```

Apply the same pattern in `google-docs/tools.go` (`readFileForWrite`, return `nil, fmt.Errorf(...)` there).

decko (approved):
All six first-round findings are properly addressed. The `_ =` confinement bypass from round 2 is fixed. `ResolvePath` design is solid — EPERM-only fallback is the right approach.

**Approved.**

Two follow-up items (not blocking):

1. **`confineRead` silent `continue` on dir resolution failure** (`plugins/bugzilla/helpers.go:230`) — pre-existing behavior, not introduced by this PR. If all allowed dirs fail to resolve, error says "path escapes allowed directories" instead of the actual resolution error. Low practical risk (both dirs are set at startup), but worth improving error reporting in a follow-up for debuggability.

2. **`resolvedSessionDir` cache path untested** — plugin tests only exercise the fallback branch (`resolvedSessionDir == ""`). The init-time hot path is never tested. Consider a test that sets both `sessionDir` and `resolvedSessionDir` to cover the primary production path.

decko (approved):

sergio-correia:
Thanks, I will work on those follow-up items in a separate PR.

PR #125, opened 2026-06-26
fix(jira): linkify bare URLs in plain-text ADF conversion

### LINKED ISSUE #124: jira: Can't make URL render as clickable link.
When using the Jira plugin against a cloud instance of Jira, neither Jira Wiki Markup, ADF or Markdown render the URL as a clickable link for the following comment:
```
Merge request for rebase to 1.5.8 (c9s): https://gitlab.com/redhat/centos-stream/rpms/python-jwcrypto/-/merge_requests/19
```

rjeffman:
Extend _parse_inline_markup to detect unbracketed https?:// URLs and emit them as link-marked text nodes, stripping trailing punctuation (.,;:!) that is not part of the URL. text_to_adf now runs plain-text paragraphs through _parse_inline_markup so bare URLs in comments become clickable links in Jira Cloud.

Also remove stale # type: ignore on jenkins test attributes that ty now resolves correctly.

Assisted-by: Claude Sonnet 4.6 (1M context) <noreply@anthropic.com>

Closes #124

### DIFF
--- a/plugins/jira/helpers.py
+++ b/plugins/jira/helpers.py
@@ -52,6 +52,7 @@ _WIKI_INLINE_RE = re.compile(
     r"(?P<monospace>\{\{(.+?)\}\})"
     r"|(?P<link>\[(?P<link_text>[^]|~]*)\|(?P<link_url>[^]]+)\])"
     r"|(?P<bare_link>\[(?P<bare_url>https?://[^]]+)\])"
+    r"|(?P<plain_url>https?://[^\s<>\[\]{}*^\\]+)"
     r"|(?P<mention>\[~(?P<mention_id>[^]]+)\])"
     r"|(?P<image>!(?P<image_url>\S+?\.\S+?)!)"
     r"|(?P<bold>(?<!\w)\*(?=\S)(?P<bold_text>.+?)(?<=\S)\*(?!\w))"
@@ -62,6 +63,32 @@ _WIKI_INLINE_RE = re.compile(
     r"|(?P<sub>(?<!\w)~(?=\S)(?P<sub_text>.+?)(?<=\S)~(?!\w))"
     r"|(?P<linebreak>\\\\)"
 )
+
+_URL_RE = re.compile(r"https?://[^\s<>\"']+")
+
+
+def _linkify_urls(text):
+    """Parse text, turning bare https?:// URLs into ADF link nodes, leaving everything else as literal text."""
+    nodes = []
+    pos = 0
+    for m in _URL_RE.finditer(text):
+        if m.start() > pos:
+            nodes.append({"type": "text", "text": text[pos : m.start()]})
+        raw = m.group()
+        url = raw.rstrip(".,;:!?")
+        if url.endswith(")") and url.count("(") < url.count(")"):
+            url = url.rstrip(")")
+        suffix = raw[len(url) :]
+        if _is_safe_url(url):
+            nodes.append({"type": "text", "text": url, "marks": [{"type": "link", "attrs": {"href": url}}]})
+        else:
+            nodes.append({"type": "text", "text": url})
+        if suffix:
+            nodes.append({"type": "text", "text": suffix})
+        pos = m.end()
+    if pos < len(text):
+        nodes.append({"type": "text", "text": text[pos:]})
+    return nodes
 
 
 def http_error(status, body):
@@ -241,6 +268,18 @@ def _parse_inline_markup(text):
                 nodes.append({"type": "text", "text": url, "marks": [{"type": "link", "attrs": {"href": url}}]})
             else:
                 nodes.append({"type": "text", "text": m.group("bare_link")})
+        elif m.group("plain_url"):
+            raw = m.group("plain_url")
+            url = raw.rstrip(".,;:!?")
+            if url.endswith(")") and url.count("(") < url.count(")"):
+                url = url.rstrip(")")
+            suffix = raw[len(url) :]
+            if _is_safe_url(url):
+                nodes.append({"type": "text", "text": url, "marks": [{"type": "link", "attrs": {"href": url}}]})
+            else:
+                nodes.append({"type": "text", "text": url})
+            if suffix:
+                nodes.append({"type": "text", "text": suffix})
         elif m.group("mention"):
             nodes.append({"type": "mention", "attrs": {"id": m.group("mention_id")}})
         elif m.group("image"):
@@ -619,11 +658,11 @@ def text_to_adf(text: str | dict | None) -> dict:
     if isinstance(text, str) and len(text) <= _MAX_WIKI_PARSE_LEN and _looks_like_wiki_markup(text):
         return wiki_to_adf(text)
 
-    # Plain text — split into paragraphs
+    # Plain text — split into paragraphs, linkifying bare URLs only
     content = []
     for para in str(text).split("\n"):
         if para.strip():
-            content.append({"type": "paragraph", "content": [{"type": "text", "text": para}]})
+            content.append({"type": "paragraph", "content": _linkify_urls(para)})
         else:
             content.append({"type": "paragraph", "content": []})
 
--- a/plugins/jira/tests/test_helpers.py
+++ b/plugins/jira/tests/test_helpers.py
@@ -372,6 +372,90 @@ class TestTextToAdf:
         assert result["type"] == "doc"
         assert result["content"][0]["content"][0]["text"] == '{"key": "value"}'
 
+    def test_plain_text_with_url_becomes_link(self):
+        result = text_to_adf("See https://example.com for details")
+        nodes = result["content"][0]["content"]
+        assert nodes[0] == {"type": "text", "text": "See "}
+        assert nodes[1]["text"] == "https://example.com"
+        assert nodes[1]["marks"] == [{"type": "link", "attrs": {"href": "https://example.com"}}]
+        assert nodes[2] == {"type": "text", "text": " for details"}
+
+    def test_plain_text_url_only(self):
+        result = text_to_adf("https://gitlab.com/redhat/centos-stream/rpms/python-jwcrypto/-/merge_requests/20")
+        nodes = result["content"][0]["content"]
+        assert len(nodes) == 1
+        assert nodes[0]["text"] == "https://gitlab.com/redhat/centos-stream/rpms/python-jwcrypto/-/merge_requests/20"
+        assert nodes[0]["marks"] == [
+            {
+                "type": "link",
+                "attrs": {"href": "https://gitlab.com/redhat/centos-stream/rpms/python-jwcrypto/-/merge_requests/20"},
+            }
+        ]
+
+    def test_plain_text_url_trailing_period_stripped(self):
+        result = text_to_adf("See https://example.com.")
+        nodes = result["content"][0]["content"]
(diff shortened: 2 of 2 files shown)


plugins/jira/helpers.py
```diff
@@ -605,11 +616,11 @@ def text_to_adf(text: str | dict | None) -> dict:
     if isinstance(text, str) and len(text) <= _MAX_WIKI_PARSE_LEN and _looks_like_wiki_markup(text):
         return wiki_to_adf(text)
 
-    # Plain text — split into paragraphs
+    # Plain text — split into paragraphs, linkifying bare URLs
     content = []
     for para in str(text).split("\n"):
         if para.strip():
-            content.append({"type": "paragraph", "content": [{"type": "text", "text": para}]})
+            content.append({"type": "paragraph", "content": _parse_inline_markup(para)})
```
decko: **Breaking behavioral regression.** `_parse_inline_markup` applies ALL wiki inline patterns — bold (`*`), italic (`_`), strikethrough (`-`), monospace (`{{}}`), images (`!`), mentions (`[~]`), etc.

The `_looks_like_wiki_markup()` guard at line 605 uses block-level patterns only, by design — its docstring says: *"to avoid false positives from inline characters like * or _ in plain text."*

This call bypasses that intent. Examples that break:
- `"The value is *approximately* 42"` → "approximately" rendered bold
- `"Set _default_value_ here"` → "default_value" rendered italic
- `"Use {{.Values.image}} in Helm"` → ".Values.image" rendered monospace

**Fix:** Use a URL-only parser here:
```python
content.append({"type": "paragraph", "content": _linkify_urls(para)})
```
Where `_linkify_urls` only matches `https?://` patterns and emits everything else as literal text.

plugins/jira/helpers.py
```diff
@@ -227,6 +228,16 @@ def _parse_inline_markup(text):
...
             else:
                 nodes.append({"type": "text", "text": m.group("bare_link")})
+        elif m.group("plain_url"):
+            raw = m.group("plain_url")
+            url = raw.rstrip(".,;:!")
+            suffix = raw[len(url) :]
+            if _is_safe_url(url):
+                nodes.append({"type": "text", "text": url, "marks": [{"type": "link", "attrs": {"href": url}}]})
+            else:
+                nodes.append({"type": "text", "text": raw})
```
decko: **Suffix duplication.** `raw` includes the trailing punctuation, then `suffix` is appended again at line 240. Output: `"https://x.com.."` instead of `"https://x.com."`.

Currently dead code (`https?://` always passes `_is_safe_url`), but structurally wrong.

**Fix:** Emit `url` (not `raw`) in the else branch, or move the suffix append inside the if block.

plugins/jira/helpers.py
```diff
@@ -227,6 +228,16 @@ def _parse_inline_markup(text):
                 nodes.append({"type": "text", "text": url, "marks": [{"type": "link", "attrs": {"href": url}}]})
             else:
                 nodes.append({"type": "text", "text": m.group("bare_link")})
+        elif m.group("plain_url"):
+            raw = m.group("plain_url")
+            url = raw.rstrip(".,;:!")
```
decko: Missing `)` and `?` from the strip set. Parenthetical URLs are very common:
- `"(see https://example.com)"` → href becomes `https://example.com)` (broken)
- `"Is it https://example.com?"` → href becomes `https://example.com?`

Consider `rstrip(".,;:!?)")`.

plugins/jira/helpers.py
```diff
@@ -49,6 +49,7 @@
     r"(?P<monospace>\{\{(.+?)\}\})"
     r"|(?P<link>\[(?P<link_text>[^]|~]*)\|(?P<link_url>[^]]+)\])"
     r"|(?P<bare_link>\[(?P<bare_url>https?://[^]]+)\])"
+    r"|(?P<plain_url>https?://\S+)"
```
decko: `\S+` is overly greedy — consumes `)`, `]`, `*`, `_`, `}`, `>`, etc. In wiki mode, `*https://example.com*` matches the entire string including the closing `*`, breaking bold markup.

Consider a more restrictive character class excluding wiki delimiters:
```python
r"|(?P<plain_url>https?://[^\s<>\[\]{}*_^~!\\]+)"
```

decko (changes requested):
URL linkification concept is sound. The main issue is the implementation approach: reusing `_parse_inline_markup()` on the plain-text path applies **all** wiki inline patterns to text that `_looks_like_wiki_markup()` explicitly determined is NOT markup.

This silently reformats existing plain-text content containing `*`, `_`, `{{}}`, `!`, etc. — a behavioral regression beyond the PR's scope.

**Recommended fix:** Create a URL-only parser (e.g. `_linkify_urls(text)`) for the plain-text path instead of reusing the full wiki inline parser.

See inline comments for details.

plugins/jira/helpers.py
sergio-correia: Adding to the point about the broader scope: one concrete side effect worth calling out is `[~accountId]` mention injection.

When `_parse_inline_markup` runs on this plain-text path, any `[~something]` pattern becomes a structured ADF mention node. Jira resolves these into @-mentions and sends a notification to that user. Before this change, that only happened when `_looks_like_wiki_markup()` returned `True`.

For example, an LLM posting a comment with `results = items[~mask]` (a NumPy indexing expression) would have `[~mask]` parsed as a mention, potentially pinging someone.

The `_linkify_urls()` approach would fix this too, since it would only match URL patterns and leave everything else as literal text.

rjeffman: Fixed. Introduced `_linkify_urls()` that only matches `https?://` URL patterns and emits everything else as literal text. The plain-text path in `text_to_adf` now calls `_linkify_urls` instead of `_parse_inline_markup`.

plugins/jira/helpers.py
rjeffman: Fixed. The else branch now emits `url` instead of `raw`, so both safe and unsafe paths go through the same trailing `if suffix:` block.

plugins/jira/helpers.py
rjeffman: Fixed. Updated to `rstrip(".,;:!?)")`.

plugins/jira/helpers.py
rjeffman: Fixed. `_WIKI_INLINE_RE` now uses `[^\s<>\[\]{}*_^~!\\]+` for `plain_url`. The separate `_URL_RE` used by `_linkify_urls` in plain-text mode uses `[^\s<>"']+` — more permissive since there is no wiki syntax to guard against there.

plugins/jira/helpers.py
rjeffman: Fixed by the `_linkify_urls()` approach — it only matches URL patterns, so `[~mask]` and similar patterns are left as literal text. Added a test with `https://example.com/~testuser/content.html?some=value#partial=field` to verify that `~` inside a URL does not trigger mention parsing.

plugins/jira/helpers.py
```diff
@@ -49,6 +49,7 @@
     r"(?P<monospace>\{\{(.+?)\}\})"
     r"|(?P<link>\[(?P<link_text>[^]|~]*)\|(?P<link_url>[^]]+)\])"
     r"|(?P<bare_link>\[(?P<bare_url>https?://[^]]+)\])"
+    r"|(?P<plain_url>https?://[^\s<>\[\]{}*_^~!\\]+)"
```
decko: `_` and `~` exclusion silently truncates URLs in wiki mode:

- `https://bugzilla.redhat.com/show_bug.cgi?id=12345` → matches only `…/show`
- `https://gitlab.com/…/merge_requests/20` → matches only `…/merge`
- `https://example.com/~user/page` → matches only `https://example.com/`

The exclusion isn't needed — the italic (`_text_`) and subscript (`~text~`) patterns already have word-boundary guards (`(?<!\w)` / `(?!\w)`), so they can't fire mid-URL.

Fix: remove `_`, `~`, `!` from the exclusion set:
```python
r"|(?P<plain_url>https?://[^\s<>\[\]{}*^\\]+)"
```

decko (changes requested):
First-round fixes look correct — `_linkify_urls()` approach, suffix fix, strip set, all good.

One blocker remains: the `plain_url` character class in `_WIKI_INLINE_RE` excludes `_` and `~`, which silently truncates common URLs in wiki mode. See inline comment.

plugins/jira/helpers.py
rjeffman: Fixed.

decko (commented):
All previously requested changes are correctly addressed. One new blocker surfaced: rstrip(".,;:!?)") unconditionally strips ), which corrupts URLs with balanced parentheses. This affects both _linkify_urls and the plain_url branch — see inline comments for details and a suggested fix.

plugins/jira/helpers.py
```diff
@@ -63,6 +64,30 @@
...
+
+def _linkify_urls(text):
+    """Parse text, turning bare https?:// URLs into ADF link nodes, leaving everything else as literal text."""
+    nodes = []
+    pos = 0
+    for m in _URL_RE.finditer(text):
+        if m.start() > pos:
+            nodes.append({"type": "text", "text": text[pos : m.start()]})
+        raw = m.group()
+        url = raw.rstrip(".,;:!?)")
```
decko, line 78: `rstrip(".,;:!?)")` strips `)` unconditionally. This breaks URLs with balanced parentheses — common in JQL filter links (`…?jql=project+in+(FOO,BAR)`) and Wikipedia URLs (`…/wiki/Foo_(bar)`): the closing `)` is stripped, leaving a dangling `(` in the `href`.

Suggested fix:
```python
url = raw.rstrip(".,;:!?")
if url.endswith(")") and url.count("(") < url.count(")"):
    url = url.rstrip(")")
```

Same fix needed in the `plain_url` branch of `_parse_inline_markup` below.

plugins/jira/helpers.py
```diff
@@ -241,6 +266,16 @@ def _parse_inline_markup(text):
                 nodes.append({"type": "text", "text": url, "marks": [{"type": "link", "attrs": {"href": url}}]})
             else:
                 nodes.append({"type": "text", "text": m.group("bare_link")})
+        elif m.group("plain_url"):
+            raw = m.group("plain_url")
```
decko, line 270: Same unbalanced-paren issue as in `_linkify_urls`. See comment above.

plugins/jira/helpers.py
rjeffman: Fixed.

plugins/jira/helpers.py
rjeffman: fixed.

decko (approved):
Blocker fixed correctly — balanced-paren logic is solid and tests cover JQL, Wikipedia, and multi-group cases.

**nit (non-blocking):** The url/suffix stripping logic is now copy-pasted verbatim in `_linkify_urls` and the `plain_url` branch of `_parse_inline_markup`. A small helper would keep future fixes in one place:
```python
def _strip_url_punctuation(raw):
    url = raw.rstrip(".,;:!?")
    if url.endswith(")") and url.count("(") < url.count(")"):
        url = url.rstrip(")")
    return url, raw[len(url):]
```

Issue #107, opened 2026-06-16
jira_get_issues returns raw ADF for comment bodies instead of plain text


decko:
## Description

When fetching issue comments via `jira_get_issues` with `fields=comment`, the comment body is returned as raw Atlassian Document Format (ADF) nodes instead of being converted to plain text or markdown.

## Expected behavior

Comment bodies should be rendered as plain text or markdown, consistent with how `jira_get_issues` renders the issue `description` field. The TOON encoding should flatten ADF into readable text.

## Actual behavior

The comment body contains raw ADF structure with `content[N]`, `type: heading`, `type: text`, `marks[1]{type}: strong`, etc. This makes it very difficult for the LLM to extract the actual comment content and wastes tokens on structural markup.

Example of what's returned:

```
body:
  content[12]:
    - attrs:
        level: 3
      content[1]{text,type}:
        Some heading text,text
      type: heading
    - content[3]:
      - text: "Some paragraph text "
        type: text
      - marks[1]{type}:
        strong
        text: "bold text"
        type: text
      - text: ", more text."
        type: text
      type: paragraph
```

## Expected output

Something like:

```
body: |
  ### Some heading text
  Some paragraph text **bold text**, more text.
```

## Reproduction

```
jira_get_issues(issue_keys="ANY-123", brief=false, fields="comment")
```

Any issue with comments in Jira Cloud (which uses ADF for comment bodies) will reproduce this.

## Notes

- The `description` field appears to be handled correctly in brief mode, so the ADF-to-text conversion logic may already exist but isn't applied to comment bodies.
- This affects `jira_get_issues` — haven't tested `jira_get_comments` separately.
- Jira Cloud stores all comment bodies in ADF format, so this affects all Cloud instances.

PR #129, opened 2026-06-30
fix(jira): set_text_field handle array fields

fullsushidev:
## Problem

`jira_set_text_field` sent values as-is to the Jira API, which fails for fields that expect `[{"name": ...}]` arrays. For example, `fixVersions` expects `[{"name": "Acme v2.7"}]` but received the plain string `"Acme v2.7"`, causing the API to return "data was not an array."

No dedicated tools exist for `fixVersions`, `versions`, or `affectsVersions`, so callers had no working path for these fields. An LLM could also pass `field_name="components"` here instead of using the dedicated `jira_set_components` tool, causing the same silent failure.

## Solution

Detect known name-array fields and auto-wrap values to `[{"name": ...}]` before the PUT call:

- **String input** (`"Acme v2.7"`) → `[{"name": "Acme v2.7"}]`
- **Empty string** (`""`) → `[]` (clears the field)
- **List of strings** (`["v1", "v2"]`) → normalized via `normalize_components` (dicts pass through unchanged)

Converted fields: `fixVersions`, `versions`, `affectsVersions`, `components`. Scalar fields like `summary` are unchanged.

## Details

- Conversion runs before the dry-run check so previews reflect the actual payload sent to Jira.
- `components` is included as a safety net — callers are steered toward the dedicated `jira_set_components` tool via the updated tool description.
- Updated `plugin.yaml` description to list supported array fields and mention the dedicated component/label tools.

<br>
Assisted-by: Claude Opus 4.6

### DIFF
--- a/plugins/jira/tests/test_tools_write.py
+++ b/plugins/jira/tests/test_tools_write.py
@@ -359,6 +359,38 @@ class TestSetTextField:
         )
         assert "name" in result["value_preview"]
 
+    def test_affectsversions_string_to_name_array(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "affectsVersions", "value": "1.0", "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["affectsVersions"] == [{"name": "1.0"}]
+
+    def test_fixversions_list_of_strings_normalized(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "fixVersions", "value": ["v1", "v2"], "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["fixVersions"] == [{"name": "v1"}, {"name": "v2"}]
+
+    def test_components_string_normalized(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "components", "value": "Web", "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["components"] == [{"name": "Web"}]
+
+    def test_components_empty_string_sends_empty_list(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "components", "value": "", "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["components"] == []
+
     def test_plain_string_field_unchanged(self):
         with _mock_http(204, {}) as mock_http:
             tools_write.set_text_field(
--- a/plugins/jira/tests/test_tools_write.py
+++ b/plugins/jira/tests/test_tools_write.py
@@ -321,6 +321,52 @@ class TestSetTextField:
             call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
             assert call_body["fields"]["description"]["type"] == "doc"
 
+    def test_fixversions_string_to_name_array(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "fixVersions", "value": "Acme v2.7", "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["fixVersions"] == [{"name": "Acme v2.7"}]
+
+    def test_versions_string_to_name_array(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "versions", "value": "Acme v2.7", "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["versions"] == [{"name": "Acme v2.7"}]
+
+    def test_fixversions_empty_string_sends_empty_list(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "fixVersions", "value": "", "dry_run": False}
+            )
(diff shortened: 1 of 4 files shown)


plugins/jira/tools_write.py
```diff
@@ -263,6 +263,12 @@ def set_text_field(params):
     if handler.is_cloud and field_name == "description":
         value = text_to_adf(value)
 
+    if isinstance(value, str):
+        if field_name in ("fixVersions", "versions", "components"):
+            value = normalize_components([v.strip() for v in value.split(",")])
```
decko: **Bug: comma-in-value splits incorrectly.**

`"Release 1, Hotfix".split(",")` produces two entries instead of one. Jira version names can contain commas.

The dedicated `jira_set_components` avoids this by accepting a list param. Consider either:
- Accepting only single values here (multi-value: use dedicated tools)
- Documenting the limitation prominently in the tool description

plugins/jira/tools_write.py
```diff
@@ -263,6 +263,12 @@ def set_text_field(params):
     if handler.is_cloud and field_name == "description":
         value = text_to_adf(value)
 
+    if isinstance(value, str):
```
decko: **Bug: empty string produces invalid payload.**

`value` defaults to `""`. With `field_name="fixVersions"`, empty string splits to `[""]`, normalizes to `[{"name": ""}]` — a bogus version sent to Jira.

Add an early guard:
```python
if isinstance(value, str) and not value.strip():
    value = []
elif isinstance(value, str):
    ...
```

decko: **Design: conversion runs after dry-run return.**

The dry-run preview (line 253-260) shows the raw string, but the actual PUT sends the converted array. Move the conversion block before `if dry_run:` so the preview reflects what Jira receives.

plugins/jira/tools_write.py
decko: **Design: overlap with dedicated tools.**

`components` already has `jira_set_components`. `labels` has `jira_set_labels`/`jira_add_labels`/`jira_remove_labels`. These dedicated tools accept proper list params and avoid the comma-splitting issue.

`fixVersions`/`versions` have no dedicated tool — that's the real gap this PR fills. Consider dropping `components`/`labels` from this path.

plugins/jira/tests/test_tools_write.py
```diff
@@ -321,6 +321,46 @@ def test_cloud_description_uses_adf(self):
...
+    def test_plain_string_field_unchanged(self):
+        with _mock_http(204, {}) as mock_http:
+            tools_write.set_text_field(
+                {"issue_key": "PROJ-1", "field_name": "summary", "value": "New title", "dry_run": False}
+            )
+            call_body = mock_http.call_args[1].get("body") or mock_http.call_args[0][3]
+            assert call_body["fields"]["summary"] == "New title"
+
 
 # --- jira_set_custom_field ---
```
decko, line 404: **Gap: no test for `versions` field.**

The code handles `("fixVersions", "versions", "components")` but only `fixVersions` and `components` are tested. Add a test for `versions`.

Also missing: empty string value, whitespace-only, and value-already-a-list passthrough.

decko (changes requested):
## Review Summary

Fills a real gap for `fixVersions`/`versions` — no dedicated tools exist for those fields. The approach is sound but needs fixes before merge.

### Bugs

1. **Comma-in-value splits incorrectly.** `"Release 1, Hotfix".split(",")` → two entries instead of one. Version and component names can contain commas. The dedicated `jira_set_components` avoids this by accepting a proper list param.

2. **Empty string → invalid payload.** `value` defaults to `""`. For `fixVersions` this produces `[{"name": ""}]`, for `labels` produces `[""]`. Both will error or corrupt data.

### Design Issues

3. **Dry-run preview shows unconverted value.** Conversion runs *after* the dry-run return. User sees `"Acme v2.7"` in preview but `[{"name": "Acme v2.7"}]` is sent. Move conversion before the dry-run block so the preview reflects reality.

4. **Overlap with dedicated tools.** `components` and `labels` already have `jira_set_components`, `jira_set_labels`, `jira_add_labels`, `jira_remove_labels`. Consider handling only `fixVersions`/`versions` here (the actual gap) and letting the dedicated tools own `components`/`labels`.

### Test Gaps

5. No test for `versions` field (code handles it, tests don't).
6. Missing edge-case tests: empty string, whitespace-only, value-already-a-list passthrough.

### Related

- Issue #107 covers the read-side counterpart (raw ADF in responses).
- PR #125 also touches Jira field formatting — no conflict but worth coordinating.

decko:
Nice fix — the core conversion logic is correct for the Jira API (both Server and Cloud).

A few things to address before merge:

**1. PR description doesn't match the code**
The description claims handling for `components`, `labels`, comma-separated splitting, and `normalize_components` reuse. The code only converts `fixVersions` and `versions`. Please update the description to match reality.

**2. Add `affectsVersions` to the conversion**
Same `[{"name": ...}]` format, no dedicated tool for it. An LLM calling `set_text_field(field_name="affectsVersions", value="1.0")` would get a Jira error.

**3. Handle list-of-strings input**
If `value = ["v1", "v2"]` (strings, not dicts), the `isinstance(value, str)` check is `False` and raw strings hit the API. `set_components` already handles this via `normalize_components` — same pattern should apply here.

**4. Guard `components` in `set_text_field`**
Nothing prevents an LLM from passing `field_name="components"` here instead of using `set_components`. It would silently fail. Consider either converting it or documenting the exclusion in the tool description.

---

For future follow-up (not blocking this PR):
- Extract a shared `normalize_name_array()` helper to DRY up the `[{"name": ...}]` pattern across version and component fields.
- Add comma-separated value splitting if multi-value input is a real use case.
- Improve dry-run preview for converted values (currently shows Python repr).

sergio-correia:
Please, remember to run the pre-commit hooks and fix the issues it points out.

fullsushidev:
All review comments have been addressed in 4ba2a7e. The remaining `ty` warnings in pre-commit are pre-existing and unrelated to this PR.

rjeffman:
LGTM. I'll still wait for @decko review.

rjeffman (approved):

decko (approved):
Reviewed with Jira and Python specialist agents. Findings:

**Medium (non-blocking):**
- `None` value unguarded for array fields — `params.get("value", "")` makes normal path safe, but explicit `None` falls through. Consider a guard or test.
- ADF `isinstance` guard is redundant (`field_name == "description"` already gates it) — harmless but adds confusion.

**Low/nit:**
- `type: string` in plugin.yaml vs "string or list" in description — minor mismatch for LLM callers.
- `type(None)` in isinstance is non-idiomatic; prefer `value is None or isinstance(...)`.

No critical issues. Jira API usage is correct, field coverage is complete, `normalize_components` reuse is appropriate, and 12 tests cover the key paths well.

Note: PR #125 (ADF linkify) also touches `tools_write.py` — check for merge conflicts when landing both.

PR #138, opened 2026-07-02
auth: add github_app auth type for GitHub App installation tokens

### LINKED ISSUE #126: auth: add github_app auth type for GitHub App installation tokens
## Proposal

Add a `github_app` auth type to wtmcp core, following the same pattern as the existing `bearer`, `basic`, `oauth2`, and `kerberos` auth providers. This enables plugins to authenticate with GitHub App installation tokens instead of personal access tokens.

### Usage

```yaml
services:
  auth:
    type: github_app
    app_id: "${GITHUB_APP_ID}"
    installation_id: "${GITHUB_INSTALLATION_ID}"
    private_key: "${GITHUB_APP_PRIVATE_KEY}"
  http:
    base_url: "${GITHUB_URL:-https://api.github.com}"
```

### What the core would handle

1. **JWT signing**: RS256 with the app's private key, 10-minute validity
2. **Token exchange**: `POST /app/installations/{installation_id}/access_tokens`
3. **Caching**: Cache token until 5 minutes before expiry
4. **Auto-refresh**: Transparent refresh on expiry (same pattern as existing `oauth2` token refresh)
5. **Bearer injection**: Standard `Authorization: Bearer <token>` header

### Why this is useful beyond a single plugin

GitHub App installation tokens are scoped per-installation, expire after 1 hour, and show the App as the actor in audit trails. This is the recommended auth method for GitHub integrations. Supporting it in wtmcp core allows any GitHub plugin to use App auth instead of long-lived PATs.

### Prior art

The existing `oauth2` auth type already handles token refresh and caching. The `github_app` provider follows the same pattern with a different grant flow (JWT assertion instead of client credentials).

decko:
## Summary

- Add `github_app` auth provider that authenticates with GitHub App installation tokens via JWT signing (RS256) and token exchange
- Provider creates short-lived JWTs, exchanges them at `POST /app/installations/{id}/access_tokens`, and caches tokens in memory with 90% lifetime refresh
- Private key loading supports both file-based (with symlink rejection + permission checks) and env var sources
- Full config plumbing through `AuthServiceConfig` → `SingleAuthConfig` → `providerFromConfig()` → `resolveAuth()`

## Plugin YAML usage

```yaml
services:
  auth:
    type: github_app
    app_id: "${GITHUB_APP_ID}"
    installation_id: "${GITHUB_INSTALLATION_ID}"
    private_key_file: "github-app.pem"
    # OR: private_key: "${GITHUB_APP_PRIVATE_KEY}"
  http:
    base_url: "${GITHUB_URL:-https://api.github.com}"
```

## Test plan

- [x] Constructor validation (empty IDs, nil transport, invalid PEM, non-RSA PEM, HTTPS enforcement, trailing slash)
- [x] JWT claims correctness (iss, iat backdated 60s, exp +10m, RS256)
- [x] Successful token exchange with header validation (Accept, User-Agent, X-GitHub-Api-Version)
- [x] Token caching and auto-refresh at 90% lifetime
- [x] HTTP error responses (401/403/404/422) with GitHub message extraction
- [x] Empty token, missing/unparseable expires_at, malformed JSON, oversized response
- [x] Private key file loading (permissions 0600, symlink rejection, file-takes-precedence)
- [x] PKCS1 and PKCS8 key format support
- [x] Concurrent access (20 goroutines, mutex correctness)
- [x] Context cancellation
- [x] All tests pass with `-race`

Closes #126


internal/auth/github_app.go
```diff
@@ -0,0 +1,255 @@
...
+		if err != nil {
+			return fmt.Errorf("github_app: parse expires_at %q: %w", tok.ExpiresAt, err)
+		}
+	}
+
+	// Refresh at 90% of the remaining lifetime to avoid edge-case failures.
+	now := time.Now()
+	remaining := expiresAt.Sub(now)
+	g.accessToken = tok.Token
+	g.expiry = now.Add(time.Duration(float64(remaining) * 0.9))
```
sergio-correia, line 210: When `expiresAt` is in the past (clock skew between the wtmcp host and GitHub/GHES, or a stale response), `remaining` goes negative and `g.expiry` is set before `now`. Every subsequent `Authenticate()` call triggers a new token exchange, creating a tight loop that exhausts GitHub's rate limits.

`RefreshTokenProvider` has a similar guard for non-positive lifetimes at `refresh_token.go:164-168`. Should have the equivalent floor here:

```suggestion
	// Refresh at 90% of the remaining lifetime to avoid edge-case failures.
	now := time.Now()
	remaining := expiresAt.Sub(now)
	if remaining < time.Minute {
		remaining = 5 * time.Minute
	}
	g.accessToken = tok.Token
	g.expiry = now.Add(time.Duration(float64(remaining) * 0.9))
```

internal/auth/github_app.go
```diff
@@ -0,0 +1,255 @@
...
+		baseURL:        baseURL,
+		client:         &http.Client{Timeout: 30 * time.Second, Transport: transport},
+	}, nil
+}
+
+// parseRSAPrivateKey decodes a PEM block and tries PKCS1 then PKCS8.
+func parseRSAPrivateKey(pemBytes []byte) (*rsa.PrivateKey, error) {
+	block, _ := pem.Decode(pemBytes)
+	if block == nil {
+		return nil, fmt.Errorf("no PEM block found in private key")
```
sergio-correia: `pem.Decode()` allocates a separate `[]byte` in `block.Bytes` with the DER-encoded key. The PEM input is zeroed via `vault.ZeroBytes`, but this intermediate DER copy stays on the heap. In a core dump the private key is recoverable from those bytes.

```suggestion
	block, _ := pem.Decode(pemBytes)
	if block == nil {
		return nil, fmt.Errorf("no PEM block found in private key")
	}
	defer vault.ZeroBytes(block.Bytes)
```

internal/auth/github_app.go
```diff
@@ -0,0 +1,255 @@
...
+
+// refreshLocked exchanges a JWT for an installation access token.
+// Must be called with g.mu held.
+func (g *GitHubAppProvider) refreshLocked(ctx context.Context) error {
+	jwtToken, err := g.createJWT()
+	if err != nil {
+		return fmt.Errorf("github_app: create JWT: %w", err)
+	}
+
+	tokenURL := fmt.Sprintf("%s/app/installations/%s/access_tokens", g.baseURL, g.installationID) //nolint:gosec // baseURL from validated config
```
sergio-correia, line 157: GitHub installation IDs are always numeric integers. Maybe validate that in the constructor so a misconfigured env var (typo, copy-paste of a different value) gets a clear error at startup rather than a confusing 404 from GitHub at runtime.

```go
if _, err := strconv.ParseInt(installationID, 10, 64); err != nil {
    return nil, fmt.Errorf("github_app: installation_id must be numeric: %s", installationID)
}
```

internal/auth/github_app_test.go
```diff
@@ -0,0 +1,756 @@
...
+	t.Helper()
+	auth := r.Header.Get("Authorization")
+	if !strings.HasPrefix(auth, "Bearer ") {
+		t.Fatalf("Authorization header = %q, want Bearer prefix", auth)
+	}
+	tokenStr := strings.TrimPrefix(auth, "Bearer ")
+
+	token, err := jwt.Parse(tokenStr, func(_ *jwt.Token) (any, error) {
+		return pubKey, nil
+	})
```
sergio-correia: The key function here accepts any signing algorithm. Since the provider explicitly uses RS256, the test should verify that. The library's key-type checking prevents algorithm confusion in practice, but an explicit check documents the requirement and catches accidental changes to the signing method.

```suggestion
	token, err := jwt.Parse(tokenStr, func(t *jwt.Token) (any, error) {
		if _, ok := t.Method.(*jwt.SigningMethodRSA); !ok {
			return nil, fmt.Errorf("unexpected signing method: %v", t.Header["alg"])
		}
		return pubKey, nil
	})
```

internal/auth/github_app_test.go
```diff
@@ -0,0 +1,756 @@
...
+	}
+
+	gap := p.(*GitHubAppProvider)
+	if gap.appID != "app-from-file" {
+		t.Errorf("appID = %q, want %q", gap.appID, "app-from-file")
+	}
+	if gap.privateKey.N.Cmp(key.N) != 0 {
+		t.Error("private key does not match file key")
+	}
+}
```
sergio-correia, line 858: The design spec lists "malformed response (non-JSON body)" and "oversized response (>1MB)" as test cases, but they're not implemented here. The malformed JSON case is worth having since corporate proxies returning HTML error pages is a common production failure. The oversized response test would protect the `LimitReader` if it's ever accidentally removed.

internal/auth/github_app.go
```diff
@@ -0,0 +1,255 @@
...
+		installationID: installationID,
+		privateKey:     key,
+		baseURL:        baseURL,
+		client:         &http.Client{Timeout: 30 * time.Second, Transport: transport},
+	}, nil
+}
+
+// parseRSAPrivateKey decodes a PEM block and tries PKCS1 then PKCS8.
+func parseRSAPrivateKey(pemBytes []byte) (*rsa.PrivateKey, error) {
+	block, _ := pem.Decode(pemBytes)
```
sergio-correia: `pem.Decode` returns a `rest` slice with any data after the first PEM block, but it's discarded here. If someone accidentally concatenates two keys into one file, this silently uses the first and ignores the second. Checking `len(bytes.TrimSpace(rest)) > 0` and returning an error would catch that kind of misconfiguration early.

sergio-correia (commented):
A few things to address, mostly around defensive edge cases.

Not in this diff, but related: `resolveAuth` in `manager.go:1453` creates the auth transport with an empty `proxy.TLSConfig{}`. For `github_app` on GHES with an internal CA configured via `services.http.tls.ca_cert`, the token exchange POST will fail with `x509: certificate signed by unknown authority` even though the HTTP proxy works fine. This is pre-existing (affects all auth providers), but `github_app` is the first provider where the token exchange target is the same host as the API, making it visible.

internal/auth/github_app.go
decko: Fixed in 89a7cd9. Added the floor with a 5-minute default (matching the RefreshTokenProvider pattern at refresh_token.go:164-168). Added TestGitHubAppNegativeExpiryFloor to verify the floor kicks in when expires_at is in the past.

internal/auth/github_app.go
decko: Fixed in 89a7cd9. Added `defer vault.ZeroBytes(block.Bytes)` right after the nil check. The parsed `*rsa.PrivateKey` still holds key material for the provider lifetime (unavoidable), but this eliminates the intermediate DER copy.

internal/auth/github_app.go
decko: Fixed in 89a7cd9. Added `strconv.ParseInt` validation in the constructor. Note: `appID` is intentionally not validated as numeric since GitHub now recommends Client IDs (format `Iv1.abc123`) for the `iss` JWT claim.

internal/auth/github_app_test.go
decko: Fixed in 89a7cd9. The `validateJWT` helper now checks `t.Method.(*jwt.SigningMethodRSA)` and rejects unexpected algorithms.

internal/auth/github_app_test.go
decko: Fixed in 89a7cd9. Added `TestGitHubAppMalformedJSON` (HTML response body), `TestGitHubAppOversizedResponse` (>1MB LimitReader cap), and `TestGitHubAppTrailingPEMData` while at it.

internal/auth/github_app.go
decko: Fixed in 89a7cd9. Now checks `len(bytes.TrimSpace(rest)) > 0` after `pem.Decode` and returns an error. Added `TestGitHubAppTrailingPEMData` which passes a doubled PEM file.

decko:
All 6 inline findings addressed in 89a7cd9:

1. **Negative lifetime floor** — added `if remaining < time.Minute { remaining = 5 * time.Minute }` with test
2. **DER bytes zeroed** — `defer vault.ZeroBytes(block.Bytes)` after PEM decode
3. **Numeric installationID validation** — `strconv.ParseInt` at construction time
4. **RS256 algorithm check in test helper** — `validateJWT` now rejects non-RSA signing methods
5. **Missing tests added** — `TestGitHubAppMalformedJSON`, `TestGitHubAppOversizedResponse`
6. **Trailing PEM data rejection** — `bytes.TrimSpace(rest)` check with test

Re: the TLS transport observation — filed as #148 since it's pre-existing and affects all auth providers.

sergio-correia (approved):
LGTM, thanks

PR #139, opened 2026-07-03
proxy(saml): fix multi-leg SSO flows for Workday-style providers

sergio-correia:
Fix SAML SSO for service providers that use multi-leg auth chains (e.g., Workday) by
following form-based redirects after the ACS POST, detecting JS redirect patterns as
SAML challenges, and scoping SSO cooldown per-URL instead of per-plugin so different
endpoints on the same service can authenticate independently.

### DIFF
--- a/internal/proxy/saml.go
+++ b/internal/proxy/saml.go
@@ -310,13 +310,16 @@ func (p *peekReadCloser) Close() error { return p.orig.Close() }
 
 // isSAMLChallenge checks if an 8KB peek from a 200 text/html
 // response looks like a SAML challenge. Fast substring pre-filter;
-// structural confirmation via parseSAMLForm happens after the full
-// body is read.
+// structural confirmation via parseSAMLForm or handleSAMLSSO happens
+// after the full body is read.
 func isSAMLChallenge(peek []byte) bool {
 	lower := bytes.ToLower(peek)
-	return bytes.Contains(lower, []byte("<form")) &&
+	if bytes.Contains(lower, []byte("<form")) &&
 		(bytes.Contains(peek, []byte("SAMLRequest")) ||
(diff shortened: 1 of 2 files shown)


decko:
Solid PR — domain validation on every chain leg, loop caps, cooldown-on-success-only, and good test quality. No blockers. A few items worth a follow-up:

**Error handling in `followSAMLChain`:**
- `body, _ = io.ReadAll(...)` silently discards read errors in both branches (form + redirect). On partial reads, next iteration parses truncated HTML. Every other `io.ReadAll` in this file checks the error — log and `break` for consistency.
- No `StatusCode` check on chain leg responses. A 4xx/5xx body gets parsed as HTML and counted as a completed leg. Both `handleSAMLSSO` and `followSAMLFormChain` check status — `followSAMLChain` should too.

**Observability:**
- When `followSAMLFormChain` fails and `handleSAMLSSO` retries, the original error is discarded. A `log.Printf("form chain failed (%v), trying redirect-based SSO", err)` before the fallback would help debugging.

**Test coverage:**
- All multi-leg tests use form-based ACS responses. The JS redirect branch of `followSAMLChain` (`<script>var redirectUrl = '...'</script>`) — the actual Workday pattern — has no end-to-end test.
- Per-URL cooldown independence (same plugin, different paths) isn't tested. Worth adding: `/path-a` triggers SSO → cooldown set → `/path-b` still triggers SSO.

**Minor:**
- `setSAMLHeaders(followReq, "")` in `followSAMLChain` POST branch passes empty Origin, unlike `handleSAMLSSO` which passes `idpBase`. SPs with CSRF protection may care.

decko (approved):

sergio-correia:
Thanks, I will work on those follow-up items in a separate PR.

PR #147, opened 2026-07-07
fix(google-docs): anchor links and numbered list rendering

### LINKED ISSUE #141: gdocs_write: anchor links render as raw markdown, numbered lists restart at 1
## Bug

`gdocs_write` with `is_markdown=true` has two conversion issues:

### 1. Anchor links render as raw markdown

`[text](#section)` appears literally as `[text](#section)` instead of being stripped or converted.

**Root cause:** `isAllowedLinkScheme` rejects `#fragment` URLs. The fallback emits `[` as a plain character, so the entire `[text](#anchor)` is parsed character-by-character and inserted verbatim.

**Expected:** Convert `#anchor` links to Google Docs internal heading links (`Link.Heading.Id`), mapping the anchor slug to the closest matching heading. If no heading matches, render just the link text (strip the markdown syntax).

### 2. Numbered lists restart at 1

Ordered list items each show `1.` when sub-bullets are present between them.

**Root cause:** `indentDepth` requires 4 spaces for nesting depth 1. Sub-bullets indented with 2-3 spaces (common in GitHub-flavored markdown) get `depth=0`, splitting the parent ordered list into separate `CreateParagraphBullets` ranges. Each range restarts numbering.

**Expected:** Accept 2+ spaces as depth 1. Items `1. A / - sub / 1. B` should render as one continuous numbered list with nested sub-bullets.

### Steps to reproduce

Write this markdown via `gdocs_write`:

```markdown
1. [Drop IOPS](#iops-right-sizing) — saves $2,560/mo
1. [Tune PostgreSQL](#postgresql-tuning) — free improvement
1. [Monitor 2-4 weeks](#decision-matrix) — then choose:
   - Option A — stay on RDS
   - Option B — migrate to Aurora
1. [NVMe / Optimized Reads](#nvme-reads) — offloads temp I/O
```

**Actual:** Raw `[text](#anchor)` in output, each item numbered `1.`
**Expected:** Clean text with heading links, sequential numbering (1, 2, 3, 4)

decko:
## Summary

- **Anchor links** (`[text](#section)`) now resolve to Google Docs internal heading bookmarks instead of rendering as raw markdown syntax
- **Numbered lists** no longer restart at 1 when sub-bullets with 2-3 space indent are present

## Changes

- `indentDepth`: accept 2+ remaining spaces as one nesting level (matches GFM conventions)
- `markdownSegment`: new `anchorSlug` field for `#fragment` links
- `parseSimpleFormattingWithDepth`: intercept `#anchor` URLs before the rejected-scheme fallback
- `mergeSegments`: prevent anchor segments from merging with adjacent plain text
- `slugifyHeading`: new GFM-compatible, unicode-safe heading slug function
- `convertMarkdownToRequests`: collect deferred anchors (new third return value)
- `resolveAnchorLinks`: new function — reads doc post-insert, maps slugs to `HeadingId`, applies `UpdateTextStyle` with `Link.Heading`; degrades gracefully on API errors

## Test plan

- [x] `TestIndentDepth` — 13 cases (0-8 spaces, tabs, mixed)
- [x] `TestOrderedListWithNestedSubBullets` — 3-space sub-bullets at depth 1
- [x] `TestParseAnchorLink` — `[text](#slug)` → anchorSlug segment
- [x] `TestParseAnchorLinkPreservesHttpLinks` — no regression for https links
- [x] `TestParseAnchorLinkRejectsOtherSchemes` — javascript: still rejected
- [x] `TestMergeSegmentsAnchorLink` — anchor segments stay separate
- [x] `TestSlugifyHeading` — 11 cases (ASCII, unicode, punctuation, edge cases)
- [x] golangci-lint clean
- [ ] Manual: write markdown with anchor links to a test Google Doc, verify heading links work

Closes #141

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/plugins/google-docs/tools.go
+++ b/plugins/google-docs/tools.go
@@ -811,7 +811,8 @@ func detectCode(doc *docs.Document) *CodeAnnotations {
 }
 
 // indentDepth converts a leading whitespace string to a nesting depth.
-// Per the markdown standard, each nesting level requires 4 spaces or 1 tab.
+// Each tab or 4-space group adds one level. A trailing group of 2-3
+// spaces also counts as one level, matching common markdown conventions.
 func indentDepth(indent string) int {
 	depth := 0
 	i := 0
@@ -824,7 +825,11 @@ func indentDepth(indent string) int {
 			depth++
 			i += 4
 		default:
-			i++
+			remaining := len(indent) - i
+			if remaining >= 2 {
+				depth++
+			}
+			i = len(indent)
 		}
 	}
 	return depth
--- a/plugins/google-docs/tools.go
+++ b/plugins/google-docs/tools.go
@@ -1727,6 +1727,10 @@ func insertMarkdownWithTables(docID string, title string, segments []markdownSeg
 			}
 		}
 
+		// Resolve anchor links now, while Phase 1 indices are still valid.
+		// Phase 2 cell population shifts post-table indices forward.
+		anchorWarnings := resolveAnchorLinks(docID, allDeferredAnchors)
+
 		// --- Phase 2: populate all table cells (chunked) ---
 		//
 		// Tables are processed in reverse document order so that cell
@@ -1749,8 +1753,6 @@ func insertMarkdownWithTables(docID string, title string, segments []markdownSeg
 			}
 			totalReplies += len(resp.Replies)
 		}
-
-		anchorWarnings := resolveAnchorLinks(docID, allDeferredAnchors)
 
 		result := map[string]any{
 			"document_id":  docID,
--- a/plugins/google-docs/tools.go
+++ b/plugins/google-docs/tools.go
@@ -1750,16 +1750,20 @@ func insertMarkdownWithTables(docID string, title string, segments []markdownSeg
 			totalReplies += len(resp.Replies)
 		}
 
-		resolveAnchorLinks(docID, allDeferredAnchors)
+		anchorWarnings := resolveAnchorLinks(docID, allDeferredAnchors)
 
-		return map[string]any{
+		result := map[string]any{
 			"document_id":  docID,
 			"title":        title,
 			"status":       "success",
 			"insert_index": insertIndex,
 			"replies":      totalReplies,
 			"tables":       tableCount,
-		}, nil
+		}
+		if len(anchorWarnings) > 0 {
+			result["unresolved_anchors"] = anchorWarnings
+		}
+		return result, nil
 	}
 
 	// No tables — single-batch insertion.
@@ -1770,16 +1774,20 @@ func insertMarkdownWithTables(docID string, title string, segments []markdownSeg
 		return nil, fmt.Errorf("batch update: %w", err)
 	}
 
-	resolveAnchorLinks(docID, deferredAnchors)
+	anchorWarnings := resolveAnchorLinks(docID, deferredAnchors)
 
-	return map[string]any{
+	result := map[string]any{
 		"document_id":  docID,
 		"title":        title,
 		"status":       "success",
 		"insert_index": insertIndex,
 		"replies":      len(resp.Replies),
 		"tables":       0,
-	}, nil
+	}
+	if len(anchorWarnings) > 0 {
+		result["unresolved_anchors"] = anchorWarnings
+	}
+	return result, nil
 }
 
 // collectTableCellRequests builds all cell-population requests for a table in
@@ -1996,8 +2004,10 @@ func populateTableCell(cell *tableCell, cellStartIndex int64) []*docs.Request {
 }
 
 type deferredAnchor struct {
-	text string
-	slug string
+	startIndex    int64
+	endIndex      int64
+	tabsAtCollect int64
+	slug          string
 }
 
 // convertMarkdownToRequests converts markdown segments to Google Docs API requests.
@@ -2529,8 +2539,10 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
 			fields = []string{"weightedFontFamily", "bold", "italic", "underline", "strikethrough"}
 			if seg.anchorSlug != "" {
 				deferredAnchors = append(deferredAnchors, deferredAnchor{
-					text: seg.text,
-					slug: seg.anchorSlug,
+					startIndex:    currentIndex,
+					endIndex:      endIndex,
+					tabsAtCollect: tabsInserted,
+					slug:          seg.anchorSlug,
 				})
 			} else if seg.linkURL != "" {
 				textStyle.Link = &docs.Link{
@@ -2548,8 +2560,10 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
 
 			if seg.anchorSlug != "" {
 				deferredAnchors = append(deferredAnchors, deferredAnchor{
-					text: seg.text,
-					slug: seg.anchorSlug,
+					startIndex:    currentIndex,
+					endIndex:      endIndex,
+					tabsAtCollect: tabsInserted,
+					slug:          seg.anchorSlug,
 				})
 			} else if seg.linkURL != "" {
 				textStyle.Link = &docs.Link{
@@ -2655,15 +2669,14 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
 	return requests, currentIndex - tabsInserted, deferredAnchors
 }
 
-func resolveAnchorLinks(docID string, anchors []deferredAnchor) {
+func resolveAnchorLinks(docID string, anchors []deferredAnchor) []string {
 	if len(anchors) == 0 {
-		return
+		return nil
 	}
 
 	doc, err := docsSvc.Documents.Get(docID).Do()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "WARNING: resolveAnchorLinks: failed to read document for heading resolution: %v\n", err)
-		return
+		return []string{fmt.Sprintf("failed to read document for heading resolution: %v", err)}
 	}
 
 	headingMap := make(map[string]string)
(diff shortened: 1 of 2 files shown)


sergio-correia (commented):
A few things to address in the anchor link resolution logic.

plugins/google-docs/tools.go
```diff
@@ -2589,7 +2651,96 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
...
+		}
+		for _, elem := range doc.Body.Content {
+			if elem.Paragraph == nil {
+				continue
+			}
+			for _, e := range elem.Paragraph.Elements {
+				if e.TextRun == nil || e.TextRun.Content == "" {
+					continue
+				}
+				idx := strings.Index(e.TextRun.Content, anchor.text)
```
sergio-correia: The text search approach here will link the wrong occurrence when the same link text appears more than once in the document. For example, two `[next section](#section-1)` and `[next section](#section-2)` links will both match the first "next section" in the doc.

Since `convertMarkdownToRequests` already tracks `currentIndex` as it builds insert requests, would it make sense to store `startIndex`/`endIndex` in `deferredAnchor` at insertion time? That way `resolveAnchorLinks` can target exact positions without searching, which also eliminates the O(M*N) scan and the `goto`.

plugins/google-docs/tools.go
```diff
@@ -2589,7 +2651,96 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
...
+}
+
+func resolveAnchorLinks(docID string, anchors []deferredAnchor) {
+	if len(anchors) == 0 {
+		return
+	}
+
+	doc, err := docsSvc.Documents.Get(docID).Do()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "WARNING: resolveAnchorLinks: failed to read document for heading resolution: %v\n", err)
```
sergio-correia: Both API errors in this function (here and at the `BatchUpdate` below) are printed to stderr and swallowed. The caller gets no indication that linking failed, and the tool returns `"status": "success"` to the user even though their anchor links were silently dropped.

Would it be better to return `error` from this function and surface a warning in the tool output? Something like an `"unresolved_links"` or `"warnings"` field in the result map would let the user know what happened.

plugins/google-docs/tools.go
```diff
@@ -2589,7 +2651,96 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
...
+		for _, e := range elem.Paragraph.Elements {
+			if e.TextRun != nil {
+				text.WriteString(e.TextRun.Content)
+			}
+		}
+		slug := slugifyHeading(strings.TrimSpace(text.String()))
+		if slug == "" {
+			continue
+		}
+		if _, exists := headingMap[slug]; !exists {
```
sergio-correia, line 2702: This gives first-wins behavior for duplicate heading slugs. If a document has two `# Results` sections, any `[link](#results)` will always point to the first one, and there is no way to target the second.

GitHub handles this by appending `-1`, `-2`, etc. to duplicate slugs. Might be worth doing the same here for compatibility, or at minimum documenting the limitation.

plugins/google-docs/tools.go
```diff
@@ -2589,7 +2651,96 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
...
+			headingMap[slug] = style.HeadingId
+		}
+	}
+
+	var requests []*docs.Request
+	for _, anchor := range anchors {
+		headingID, ok := headingMap[anchor.slug]
+		if !ok {
```
sergio-correia, line 2711: When a slug doesn't match any heading (typo, missing heading), this silently skips it. The user gets no feedback that their anchor link was unresolved.

Maybe collect the unresolved slugs and include them in the tool output as a warning? That would save the user from having to manually check each link in the doc.

plugins/google-docs/tools.go
decko: Good catch — the text-based search will match the wrong occurrence with duplicate link text. Will switch to index-based tracking with per-anchor `tabsInserted` adjustment to get exact positions. This also eliminates the O(M*N) scan and the `goto`.

plugins/google-docs/tools.go
decko: Agreed on surfacing warnings — but the function should not return `error` since failing the write after text is already inserted would be worse UX. Will change to return `[]string` (warnings) and add an `"unresolved_anchors"` field to the result map. This also covers the unresolved slug feedback below.

plugins/google-docs/tools.go
decko: First-wins was a deliberate scope decision for this PR — GFM-style `-1`/`-2` disambiguation adds complexity for a rare edge case. Would you be OK with a follow-up ticket to add duplicate slug support? We can document the limitation here and address it separately.

plugins/google-docs/tools.go
decko: Folding this into the warnings approach from the error handling comment above — unresolved slugs will be collected and returned in the result map.

plugins/google-docs/tools.go
sergio-correia: Sure, you can do this in a follow-up ticket, no problem.

plugins/google-docs/tools.go
```diff
@@ -1441,6 +1456,28 @@ func isAllowedLinkScheme(rawURL string) bool {
 		strings.HasPrefix(lower, "mailto:")
 }
 
+// slugifyHeading produces a GFM-compatible anchor slug from a heading string.
+// It lowercases, keeps letters/digits/spaces/hyphens, converts spaces to
+// hyphens, and trims leading/trailing hyphens.
+func slugifyHeading(text string) string {
```
rjeffman, line 1463: 1. slugifyHeading drops underscores, breaking GFM compatibility (correctness)

  The PR description claims GFM-compatible slugification, but GFM preserves underscores in heading slugs. The current
  implementation silently strips them because underscores match none of the three cases:

  case unicode.IsLetter(r) || unicode.IsDigit(r):   // underscore is not a letter or digit
  case r == ' ' || r == '-':                          // underscore is not space or hyphen
  // default: dropped

  A heading ## My_Function_Name produces slug myfunctionname instead of GFM's my_function_name. Any [link](#my_function_name)
  would fail to resolve.

  Fix:
  case unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_':
      b.WriteRune(r)
      prevHyphen = false

plugins/google-docs/tools.go
```diff
@@ -2590,7 +2665,81 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
...
+
+func resolveAnchorLinks(docID string, anchors []deferredAnchor) []string {
+	if len(anchors) == 0 {
+		return nil
+	}
+
+	doc, err := docsSvc.Documents.Get(docID).Do()
+	if err != nil {
+		return []string{fmt.Sprintf("failed to read document for heading resolution: %v", err)}
+	}
```
rjeffman, line 2677: 2. Nil dereference risk in resolveAnchorLinks if doc.Body is nil (robustness)

  doc, err := docsSvc.Documents.Get(docID).Do()
  if err != nil {
      return []string{...}
  }
  // No nil check on doc.Body before:
  for _, elem := range doc.Body.Content {

  If the API returns a document with a nil Body (empty document, partial response, API quirk), this panics. Since this
  function already degrades gracefully on errors (returning warnings instead of propagating errors), a nil body should follow
  the same pattern:

  if doc.Body == nil {
      return []string{"document body is empty, cannot resolve heading anchors"}
  }

plugins/google-docs/tools.go
```diff
@@ -2489,7 +2557,14 @@ func convertMarkdownToRequests(segments []markdownSegment, startIndex int64, str
 				Strikethrough: seg.strikethrough,
 			}
 
-			if seg.linkURL != "" {
+			if seg.anchorSlug != "" {
```
rjeffman: This is duplicated in 2539. It should be moved to line 2526, and only `if seg.linkURL != ""` be handled here.

rjeffman (commented):
Have you tested with tables in the document?

LLM security reviewer spotted this issue:
```
  4. Index desynchronization risk with tables (data integrity, medium security)
  Deferred anchor indices are captured during text construction, but resolution happens after table insertion and formatting
  passes may have shifted indices. If indices drift, UpdateTextStyle applies links to wrong text ranges — silent document
  corruption. The security reviewer flagged this as the most subtle issue in the PR.
```

plugins/google-docs/tools.go
decko: Will create a follow-up ticket for duplicate slug disambiguation.

plugins/google-docs/tools.go
decko: Fixed in 31b26d9 — added `|| r == '_'` to the keep condition. Test case added.

plugins/google-docs/tools.go
decko: Fixed in 31b26d9 — added nil guard on `doc.Body` before iterating.

plugins/google-docs/tools.go
decko: Fixed in 31b26d9 — hoisted `anchorSlug` collection above the `isInlineCode` branch.

sergio-correia:
> Have you tested with tables in the document?
> 
> LLM security reviewer spotted this issue:
> 
> ```
>   4. Index desynchronization risk with tables (data integrity, medium security)
>   Deferred anchor indices are captured during text construction, but resolution happens after table insertion and formatting
>   passes may have shifted indices. If indices drift, UpdateTextStyle applies links to wrong text ranges — silent document
>   corruption. The security reviewer flagged this as the most subtle issue in the PR.
> ```

LGTM, but I would like to know the answer to this question from @rjeffman

decko (commented):
Good catch — the index desynchronization with tables is a real bug.

**Confirmed via live testing:** with the original code, anchor links placed after a table get their indices shifted by the content inserted during Phase 2 (table cell population). The result is silent document corruption:
- Heading links applied to wrong text inside table cells (e.g., "Cell 7" gets an unwanted heading link)
- Partial link application on wrong words after the table (e.g., "that should poi" gets linked instead of "anchor link")
- Second anchor link dropped entirely

**Root cause:** `resolveAnchorLinks` was called *after* Phase 2 cell population, but anchor indices were captured during Phase 1. Cell population shifts all post-table indices forward, making the captured indices stale.

**Fix (19b00d9):** Move `resolveAnchorLinks` from after Phase 2 to after Phase 1 (text + empty table skeletons). At that point all body text is inserted and indices are still valid. `UpdateTextStyle` (link-only, no text insertion) doesn't disturb Phase 2 cell indices.

Verified with a live test doc — both anchor links now land on the correct text, no stray links in table cells.

sergio-correia:
@decko, thanks now it looks good to me, but please rebase on top of the current tree.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/LeGambiArt/wtmcp
 
-go 1.26.6
+go 1.26.4
```
sergio-correia: Please drop this change and run `make tidy` to make sure this file will be up to date.

sergio-correia (approved):
LGTM

PR #176, opened 2026-07-22
feat: add PR/MR creation tools for GitHub and GitLab

Cropi:
Add `github_create_pull_request` to the GitHub plugin and `gitlab_create_merge_request` to the GitLab plugin.

Both tools default to `dry_run=true` for safety.

### GitHub — `github_create_pull_request`

Required: `repo`, `title`, `head`, `base`
Optional: `body`, `draft`, `maintainer_can_modify`, `dry_run`

### GitLab — `gitlab_create_merge_request`

Required: `project_id`, `source_branch`, `target_branch`, `title`
Optional: `instance`, `description`, `labels`, `remove_source_branch`, `squash`, `draft` (prepends `Draft: ` unless already prefixed), `dry_run`

### Testing
- Added unit tests for both plugins covering the new write tools

### DIFF
--- a/plugins/github/tools_write.py
+++ b/plugins/github/tools_write.py
@@ -277,8 +277,77 @@ def add_comment(params):
     return result
 
 
+def create_pull_request(params):
+    """Create a new pull request."""
+    repo = params.get("repo", "")
+    title = params.get("title", "")
+    head = params.get("head", "")
+    base = params.get("base", "")
+    body = params.get("body", "")
+    draft = params.get("draft", False)
+    maintainer_can_modify = params.get("maintainer_can_modify", True)
+    dry_run = params.get("dry_run", True)
+
+    _validate_repo(repo)
+    owner, name = _split_repo(repo)
+
+    if not title or not title.strip():
+        raise ValueError("title is required and must be non-empty")
+    if not head or not head.strip():
+        raise ValueError("head is required and must be non-empty")
+    if not base or not base.strip():
+        raise ValueError("base is required and must be non-empty")
+    if body and len(body) > _MAX_BODY_LEN:
+        raise ValueError(f"body exceeds {_MAX_BODY_LEN} character limit ({len(body)} chars)")
+
+    if dry_run:
+        preview = {
+            "dry_run": True,
+            "action": "github_create_pull_request",
+            "repo": repo,
+            "head": head,
+            "base": base,
+            "title": title,
+            "draft": draft,
+            "maintainer_can_modify": maintainer_can_modify,
+        }
+        if body:
+            preview["body_preview"] = body[:200]
+        return preview
+
+    api_body = {
+        "title": title,
+        "head": head,
+        "base": base,
+        "draft": draft,
+        "maintainer_can_modify": maintainer_can_modify,
+    }
+    if body:
+        api_body["body"] = body
+
+    status, resp, headers = handler.http(
+        "POST",
+        f"/repos/{owner}/{name}/pulls",
+        body=api_body,
+    )
+    if status < 200 or status >= 300:
+        return _http_error(status, resp)
+
+    handler.invalidate_cache()
+    result = {
+        "id": resp.get("id") if isinstance(resp, dict) else None,
+        "number": resp.get("number") if isinstance(resp, dict) else None,
+        "html_url": resp.get("html_url") if isinstance(resp, dict) else None,
+        "state": resp.get("state") if isinstance(resp, dict) else None,
+        "draft": resp.get("draft") if isinstance(resp, dict) else None,
+    }
+    _check_rate_limit(headers, result)
+    return result
+
+
 WRITE_TOOLS = {
     "github_create_review": create_review,
     "github_add_pr_comment": add_pr_comment,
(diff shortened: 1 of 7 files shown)


plugins/github/tools_write.py
```diff
@@ -277,8 +277,77 @@ def add_comment(params):
...
+    if not title or not title.strip():
+        raise ValueError("title is required and must be non-empty")
+    if not head or not head.strip():
+        raise ValueError("head is required and must be non-empty")
+    if not base or not base.strip():
+        raise ValueError("base is required and must be non-empty")
+    if body and len(body) > _MAX_BODY_LEN:
+        raise ValueError(f"body exceeds {_MAX_BODY_LEN} character limit ({len(body)} chars)")
+
+    if dry_run:
```
sergio-correia: When `dry_run` comes in as JSON `null`, Python deserializes it to `None`. Since `params.get("dry_run", True)` finds the key (it exists, just null), it returns `None` rather than the default `True`. Then `if None:` is falsy, so the dry_run gate is skipped and a real PR gets created.

The Go side handles this correctly with `isDryRun(v *bool) bool { return v == nil || *v }`, treating null as "yes, dry run."

Would it be better to flip the condition here?

```suggestion
    if dry_run is not False:
```

That way only an explicit `False` disables the safety gate, matching the Go semantics. The same pattern applies to `create_review`, `add_pr_comment`, and `add_comment` above (pre-existing).

plugins/gitlab/tools_write.go
```diff
@@ -191,6 +191,111 @@ func toolAddMRNote(params, _ json.RawMessage) (any, error) {
...
+		return nil, fmt.Errorf("description exceeds %d character limit (%d chars)", maxBodyLen, len([]rune(p.Description)))
+	}
+
+	title := p.Title
+	if p.Draft && !strings.HasPrefix(p.Title, "Draft: ") {
+		title = "Draft: " + title
+	}
+
+	if isDryRun(p.DryRun) {
+		m := map[string]any{
```
sergio-correia, line 242: The preview map includes `project_id`, `source_branch`, `target_branch`, `title`, and optionally `description_preview`/`labels`, but it omits `remove_source_branch`, `squash`, and `draft` (as an explicit boolean). A user previewing an MR creation with `remove_source_branch: true, squash: true` gets no indication those options were received.

The GitHub `create_pull_request` preview does include `draft` and `maintainer_can_modify` for full visibility. Would it make sense to add the missing fields here for consistency?

```go
if p.Draft {
    m["draft"] = true
}
if p.RemoveSourceBranch != nil {
    m["remove_source_branch"] = *p.RemoveSourceBranch
}
if p.Squash != nil {
    m["squash"] = *p.Squash
}
```

plugins/gitlab/tools_write.go
```diff
@@ -191,6 +191,111 @@ func toolAddMRNote(params, _ json.RawMessage) (any, error) {
...
+			}
+			m["description_preview"] = string(runes)
+		}
+		if len(p.Labels) > 0 {
+			m["labels"] = p.Labels
+		}
+		return m, nil
+	}
+
+	client, err := resolveInstance(p.Instance)
```
sergio-correia: `resolveInstance` is called here, after the `isDryRun` check returns early on line 236. This means a dry-run preview succeeds even with a bogus `instance` name (e.g., a typo like `"stagingg"`). The user only discovers the misconfiguration when they set `dry_run=false`.

`toolCreateMRDiscussion` calls `resolveInstance` before the dry-run check (because it needs to fetch diff refs), so it catches bad instance names eagerly. Have you considered moving this call above `isDryRun` here too, so the preview validates everything it can?

plugins/gitlab/tools_write.go
```diff
@@ -191,6 +191,111 @@ func toolAddMRNote(params, _ json.RawMessage) (any, error) {
...
+	}
+	if strings.TrimSpace(p.Title) == "" {
+		return nil, fmt.Errorf("title is required")
+	}
+	if len([]rune(p.Description)) > maxBodyLen {
+		return nil, fmt.Errorf("description exceeds %d character limit (%d chars)", maxBodyLen, len([]rune(p.Description)))
+	}
+
+	title := p.Title
+	if p.Draft && !strings.HasPrefix(p.Title, "Draft: ") {
```
sergio-correia: This prefix check is case-sensitive and only handles `"Draft: "`. GitLab also recognizes `"draft: "` (lowercase) and the legacy `"WIP: "` as draft indicators. If someone passes `title: "draft: fix login"` with `draft: true`, the result is `"Draft: draft: fix login"`.

Have you considered a case-insensitive check here?

```suggestion
	if p.Draft && !strings.HasPrefix(strings.ToLower(p.Title), "draft: ") {
```

sergio-correia (commented):
A few issues to address before merging.

plugins/github/tools_write.py
Cropi: Makes sense, I modified the other pre-existing ones as well.

sergio-correia (approved):
LGTM, thanks!

PR #177, opened 2026-07-22
build: add make nosandbox target

walac:
Building without the arapuca sandbox previously required knowing the raw go build invocation with -tags nosandbox. Several documentation files also referenced stale target names like build-nosandbox and build-sandbox that do not exist in the current Makefile.

Add a nosandbox target that builds both wtmcp and wtmcpctl with the nosandbox build tag, depending on plugins but skipping the arapuca dependency. Update references in BUILDING.md, CLAUDE.md, README.md, the plugin guide, and the config.go deprecation warning to use the new target name consistently.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -32,7 +32,8 @@ from the `third_party/arapuca` submodule (requires a Rust toolchain).
 To build without sandbox support:
 
 ```bash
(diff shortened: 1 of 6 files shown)


sergio-correia (approved):
LGTM

PR #178, opened 2026-07-23
Add read/write tools to GitHub and GitLab plugins, harden existing handlers

sergio-correia:
New tool handlers for GitHub (7 read, 2 write), GitLab (3 read, 2 write), and Testing Farm
  (1 cooperative poll), plus core manifest schema improvements and hardening fixes across all
  four plugins.

  - GitHub/GitLab write tools: create_branch and create_or_update_file with branch name
  validation, dry_run defaults, and path traversal rejection
  - GitHub/GitLab read tools: file contents, code/repo search, PR listing, commit listing -
  with caching, truncation, and PII-minimized output
  - Testing Farm: wait_for_result cooperative poll tool with transient error handling and
  call-budget awareness
  - Core manifest: auto-fix array params missing items definition; extend ItemsDef with
  Properties/Required for nested object schemas
  - Hardening: nil-check MR Author, dry_run is not False guard, ApiError consistency, cache
  invalidation logging, parseTime error returns, MaxResults/MaxFiles bounds checks, trace
  read limiting, assignee_username wiring via user ID resolution

### DIFF
--- a/plugins/github/tools.py
+++ b/plugins/github/tools.py
@@ -40,11 +40,15 @@ def _check_rate_limit(headers, result):
 def _validate_org(org):
     if org and not _VALID_ORG.match(org):
         raise ValueError(f"invalid org name: {org!r}")
+    if org == "..":
+        raise ValueError(f"invalid org name: {org!r}")
 
 
 def _validate_repo(repo):
     if not _VALID_REPO.match(repo):
         raise ValueError(f"invalid repo format: {repo!r} (expected owner/repo)")
+    if ".." in repo.split("/"):
+        raise ValueError(f"invalid repo format: {repo!r} (path traversal)")
 
 
 def _validate_path(path):
--- a/plugins/github/tools.py
+++ b/plugins/github/tools.py
@@ -1,5 +1,6 @@
 """GitHub tool implementations."""
 
+import base64
 import hashlib
 import re
 
@@ -8,7 +9,10 @@ import handler
 _VALID_ORG = re.compile(r"^[a-zA-Z0-9._-]+$")
 _VALID_REPO = re.compile(r"^[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+$")
 _VALID_FILTER = {"assigned", "created", "mentioned", "subscribed", "repos", "all"}
+_VALID_PR_STATE = {"open", "closed", "all"}
 _RATE_LIMIT_WARN = 100
+_MAX_CONTENT_BYTES = 50000
+_MAX_DIR_ENTRIES = 200
 
 
 def _http_error(status, body):
@@ -41,6 +45,15 @@ def _validate_org(org):
 def _validate_repo(repo):
     if not _VALID_REPO.match(repo):
         raise ValueError(f"invalid repo format: {repo!r} (expected owner/repo)")
+
+
+def _validate_path(path):
+    if not path or not path.strip():
+        raise ValueError("path is required")
+    if ".." in path.split("/"):
+        raise ValueError(f"path traversal in path: {path!r}")
+    if path.startswith("/"):
+        raise ValueError(f"absolute path not allowed: {path!r}")
 
 
 def _validate_filter(f):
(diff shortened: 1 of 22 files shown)


decko (commented):
## PR #178 Review — Findings Summary

Reviewed by a specialist team (Security, Go, Python, Adversarial QE). All findings below are **new code introduced by this PR** — pre-existing issues excluded.

### Critical

| ID | Finding |
|----|---------|
| **C1** | **Path traversal in GitHub file tools via `path` param.** `path` is interpolated into API URLs without traversal checks. Go's `url.JoinPath` resolves `..`, so `path="../../../../repos/victim/private/contents/secret.env"` escapes the intended repo. Affects `get_file_contents` (read) and `create_or_update_file` (write). GitLab is NOT affected (go-gitlab uses `PathEscape`). |

### High

| ID | Finding |
|----|---------|
| **H1** | **`wait_for_result` timeout/budget exit paths can raise unhandled `ApiError`.** The final `testing_farm_get_request()` calls on timeout/budget expiry are outside the `try/except ApiError` block. A 502 at that moment surfaces a raw exception instead of the structured "Re-invoke to continue" response. |

### Medium

| ID | Finding |
|----|---------|
| **M1** | Unbounded base64 decode in GitHub `get_file_contents` — full content decoded into memory before size check. GitLab does this correctly with `io.LimitReader`. |
| **M3** | `toolCreateMergeRequest` dry_run skips instance validation after this PR moves `resolveInstance` below the dry_run check. `toolCreateOrUpdateFile` still validates instance in dry_run — inconsistent within this PR. |
| **M4** | Param naming: `message` (GitHub) vs `commit_message` (GitLab) for the same concept in `create_or_update_file`. LLMs will cross-contaminate. |
| **M5** | `create_branch` dry_run: GitHub makes HTTP calls (validates source branch), GitLab returns immediately (silently accepts invalid branches). |
| **M6** | GitLab `get_file_contents` hardcodes `ref` default to `"main"`. GitHub lets the API decide. Repos with `master` or other defaults will get 404s. |
| **M7** | Three `wait_for_result` tests (`test_polls_until_complete`, `test_transient_error_retry`, `test_consecutive_errors_abort`) use real `time.monotonic` — flaky on slow CI. |

### Low

| ID | Finding |
|----|---------|
| **L1** | Missing branch validation in `create_or_update_file` (both plugins) — `create_branch` validates, `create_or_update_file` doesn't. |
| **L3** | No test for `timeout_minutes` clamping boundaries (0→1, 999→120). |
| **L4** | Content size limit: Python `len(str)` = characters vs Go `len(string)` = bytes — different effective limits for non-ASCII content. |
| **L5** | `ParamsSchema()` drops `Default` from nested item properties in the new Properties/Required support. |
| **L6** | `get_file_contents`: GitHub supports directory listing, GitLab does not — undocumented capability gap. |
| **L7** | `userIDCache` (`sync.Map`) grows unbounded, no TTL or eviction. |
| **L8** | Test `TestCreateOrUpdateFileDryRunNew`: `WriteHeader(404)` called before `Content-Type` header is set. |

plugins/github/tools.py
```diff
@@ -523,6 +531,238 @@ def search(params):
...
+    if ref:
+        query["ref"] = ref
+
+    cache_input = f"file:{owner}/{name}/{path}:{ref}"
+    cache_key = f"file:{hashlib.sha256(cache_input.encode()).hexdigest()[:12]}"
+    cached = handler.cache_get(cache_key)
+    if cached:
+        return cached
+
+    status, body, headers = handler.http("GET", f"/repos/{owner}/{name}/contents/{path}", query=query or None)
```
decko, line 562: **[C1 — Critical: Path traversal]** `path` is user-controlled and interpolated directly into the API URL. Go's `url.JoinPath` resolves `..` segments, so a path like `../../../../repos/victim-org/private-repo/contents/secret.env` escapes the intended repo scope.

This bypasses `_validate_repo()` entirely since traversal happens through `path`, not `repo`.

Suggested fix — reject `..` as a path component:
```python
def _validate_path(path):
    if not path or not path.strip():
        raise ValueError("path is required")
    if ".." in path.split("/"):
        raise ValueError(f"path traversal rejected: {path!r}")
```

Same issue in `tools_write.py` lines 442 and 469.

plugins/github/tools_write.py
```diff
@@ -345,9 +364,130 @@ def create_pull_request(params):
...
+        raise ValueError("message is required")
+    if len(content) > _MAX_FILE_CONTENT:
+        raise ValueError(f"content exceeds {_MAX_FILE_CONTENT} character limit ({len(content)} characters)")
+
+    query = {}
+    if branch:
+        query["ref"] = branch
+
+    existing_sha = None
+    status, body, _ = handler.http("GET", f"/repos/{owner}/{name}/contents/{path}", query=query or None)
```
decko, line 443: **[C1 — Critical: Path traversal]** Same issue as `tools.py:554`. The `path` param is interpolated into the API URL without traversal checks. Both the GET (line 442) and PUT (line 469) are affected. An attacker can write files to any repo the token can access.

Apply `_validate_path()` before any URL construction.

plugins/testing-farm/handler.py
```diff
@@ -354,6 +364,89 @@ def testing_farm_get_request(params):
...
+    deadline = start + timeout_minutes * 60
+    call_deadline = start + _CALL_BUDGET
+    poll_count = 0
+    consecutive_errors = 0
+    last_state = ""
+
+    while True:
+        now = time.monotonic()
+        if now >= deadline:
+            result = testing_farm_get_request({"request_id": request_id})
```
decko: **[H1 — High: Unhandled ApiError on timeout exit]** This `testing_farm_get_request()` call is outside the `try/except ApiError` block in the polling loop. If the API returns 502 when the deadline expires, the user gets a raw exception instead of the structured `{"waited": True, "timeout": True}` response.

Same issue on line 407 (call budget exit) and line 438 (poll success).

Suggested fix — wrap in `try/except`:
```python
try:
    result = testing_farm_get_request({"request_id": request_id})
except Exception:
    result = {"id": request_id, "state": last_state or "unknown"}
```

plugins/github/tools.py
```diff
@@ -523,6 +531,238 @@ def search(params):
...
+        "html_url": body.get("html_url", ""),
+    }
+
+    if content_b64 is None:
+        result["content_available"] = False
+        result["download_url"] = body.get("download_url", "")
+        result["git_url"] = body.get("git_url", "")
+    else:
+        try:
+            decoded = base64.b64decode(content_b64)
```
decko, line 602: **[M1 — Medium: Unbounded base64 decode]** The full `content_b64` is decoded into memory before the size check on line 591. On GHE or a compromised proxy, a multi-GB `content` field causes unbounded memory allocation.

The GitLab equivalent does this correctly with `io.LimitReader`. Consider checking the `size` field from the API response before decoding:
```python
if size > _MAX_CONTENT_BYTES:
    result["content_available"] = False
    return result
```

plugins/gitlab/tools_write.go
```diff
@@ -269,6 +287,11 @@ func toolCreateMergeRequest(params, _ json.RawMessage) (any, error) {
 		return m, nil
 	}
 
+	client, err := resolveInstance(p.Instance)
```
decko: **[M3 — Medium: Dry-run skips instance validation]** `resolveInstance` was moved here (after the `isDryRun` return on line 259), so dry_run with a bogus instance name produces a valid-looking preview.

This is a regression — the old code validated instance before the dry_run check. Also inconsistent within this PR: `toolCreateOrUpdateFile` (line 430) still calls `resolveInstance` before `isDryRun`.

Consider validating just the instance name (without creating a full client) before the dry_run return, or keep `resolveInstance` before the check.

plugins/gitlab/plugin.yaml
```diff
@@ -345,6 +412,64 @@ tools:
...
+        description: "Project ID or path (e.g. 'group/project')"
+      path:
+        type: string
+        required: true
+        description: "File path in the repository"
+      content:
+        type: string
+        required: true
+        description: "File content (plain text)"
+      commit_message:
```
decko: **[M4 — Medium: Param naming inconsistency]** This uses `commit_message` while the equivalent GitHub `create_or_update_file` uses `message` for the same concept. LLMs switching between platforms will use the wrong name.

Consider aligning to one name across both plugins.

plugins/gitlab/tools_write.go
```diff
@@ -305,6 +328,167 @@ func toolCreateMergeRequest(params, _ json.RawMessage) (any, error) {
...
+		return nil, err
+	}
+	if p.SourceBranch == "" {
+		p.SourceBranch = "main"
+	}
+	if err := validateBranch(p.SourceBranch); err != nil {
+		return nil, fmt.Errorf("source_branch: %w", err)
+	}
+
+	if isDryRun(p.DryRun) {
```
decko, line 379: **[M5 — Medium: Dry-run behavioral inconsistency]** GitLab `create_branch` returns a preview immediately in dry_run without any server call. GitHub's equivalent fetches the source branch SHA even in dry_run to validate it exists.

Result: GitHub dry_run catches invalid source branches; GitLab dry_run silently accepts them.

plugins/gitlab/tools.go
```diff
@@ -931,3 +1001,208 @@ func timeStr(t *time.Time) string {
...
+	}
+
+	client, err := resolveInstance(p.Instance)
+	if err != nil {
+		return nil, err
+	}
+
+	ref := p.Ref
+	if ref == "" {
+		ref = "main"
```
decko: **[M6 — Medium: Hardcoded ref default]** This forces `"main"` when ref is empty. GitHub's equivalent lets the API decide the default branch. Repos using `master` or other default branches will get 404s unless the user explicitly passes `ref`.

Fix: omit `ref` from the API call when empty, letting GitLab use the project's configured default branch.

plugins/testing-farm/tests/test_handler.py
```diff
@@ -835,3 +835,149 @@ def test_no_artifacts_url(self):
...
+    def test_already_complete(self):
+        """Request in terminal state on first check returns immediately."""
+        with _mock_cache_get(None), _mock_http(200, SAMPLE_REQUEST), _mock_cache_set():
+            result = handler.testing_farm_wait_for_result({"request_id": REQ_ID})
+        assert result["waited"] is True
+        assert result["completed"] is True
+        assert result["poll_count"] == 1
+        assert result["state"] == "complete"
+
+    def test_polls_until_complete(self):
```
decko, line 859: **[M7 — Medium: Flaky test — real time.monotonic]** This test (and `test_transient_error_retry` at line 927, `test_consecutive_errors_abort` at line 944) mocks `time.sleep` but not `time.monotonic`. The 45-second call budget uses wall time. On slow CI hosts, budget expiry can fire early, causing spurious failures.

`test_call_budget_exceeded` and `test_timeout` correctly mock `time.monotonic` — apply the same pattern here.

plugins/github/tools.py
sergio-correia: Fixed. Added `_validate_path()` that rejects `..` segments and absolute paths, called in both `get_file_contents` and `create_or_update_file`. Also added a size pre-check (`size > 1_000_000`) to skip base64 decode for very large files. The GitLab plugin got the equivalent `validatePath()` in Go. Tests added for both.

plugins/github/tools_write.py
sergio-correia: Fixed. `_validate_path` is imported from `tools.py` and called before any URL construction. Also added `_validate_branch(branch)` for the branch param (was missing). Test added.

plugins/testing-farm/handler.py
sergio-correia: Fixed. All three exit-path `testing_farm_get_request()` calls (timeout, call budget, terminal state) are now wrapped in `try/except Exception` with a fallback dict. If the API is unreachable at the exit moment, the function returns a structured response with `state: last_state` instead of a raw exception.

plugins/github/tools.py
sergio-correia: Fixed. Added a `size > 1_000_000` guard before base64 decode. Files over 1MB skip decoding entirely and return `content_available: False`. Files between 50KB and 1MB still get truncated display (preserving the useful first-50KB behavior). The GitLab equivalent already bounds via `io.LimitReader`.

plugins/gitlab/tools_write.go
sergio-correia: Fixed. `resolveInstance` is now called before `isDryRun` in all five write tools (`CreateMRDiscussion`, `AddMRNote`, `CreateMergeRequest`, `CreateBranch`, `CreateOrUpdateFile`). Consistent pattern throughout. A dry-run with a bogus instance now fails immediately.

plugins/gitlab/plugin.yaml
sergio-correia: Fixed. Renamed to `message` in the GitLab plugin (plugin.yaml, Go struct json tag, validation error, tests) to match GitHub's `create_or_update_file`.

plugins/gitlab/tools_write.go
sergio-correia: Acknowledged. This is an intentional design choice. GitLab's dry-run skips the server call because it doesn't need any server data for the preview. GitHub's dry-run validates the source branch exists, which is useful but costs an API call and a rate-limit token. Both approaches are valid; documenting the difference rather than forcing alignment.

plugins/gitlab/tools.go
sergio-correia: Fixed. Both `toolGetFileContents` and `toolCreateOrUpdateFile` now pass nil for the ref when empty (via `omitempty` on the `*string` field), letting the GitLab API use the project's configured default branch. Plugin.yaml descriptions updated to say "project default branch" instead of "main".

plugins/testing-farm/tests/test_handler.py
sergio-correia: Fixed. Added `patch("time.monotonic", return_value=0.0)` to `test_polls_until_complete`, `test_transient_error_retry`, and `test_consecutive_errors_abort`, matching the pattern already used by `test_call_budget_exceeded` and `test_timeout`.

decko (approved):
LGTM. All critical and high findings from the initial review have been addressed. Remaining items are non-blocking (tracked in #180, #181, #182).

Reviewed by specialist team (Security, Go, Python, Adversarial QE) across two review rounds.

PR #183, opened 2026-07-27
manifest: omit "type" from schema when YAML does not specify one

sergio-correia:
ParamsSchema() unconditionally emitted "type": "" for parameters and item properties that omit the type field in YAML. An empty string is not a valid JSON Schema type, so the jsonschema compiler rejected the schema and permanently disabled the affected tool (e.g. jira_create_issue). Guard all three emission sites to only include "type" when non-empty, which is valid JSON Schema for "accept any type".

### DIFF
--- a/internal/plugin/manifest.go
+++ b/internal/plugin/manifest.go
@@ -598,7 +598,10 @@ func (t *ToolDef) ParamsSchema() map[string]any {
 	var required []string
 
 	for name, p := range t.Params {
-		prop := map[string]any{"type": p.Type}
+		prop := make(map[string]any)
+		if p.Type != "" {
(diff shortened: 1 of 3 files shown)


decko (approved):
LGTM. Clean fix, well-tested.

Two minor non-blocking observations:

1. No negative validation test for "required + untyped" — confirming the schema still rejects a missing `value` when it's in the `required` list. Low risk since `required` enforcement is orthogonal to `type` in JSON Schema.
2. Top-level untyped param and array items are tested at schema-generation level only, not through the `CompileParamsSchema` round-trip. The item-property level is covered end-to-end.

Pre-existing issues filed separately: #184, #185.

Issue #116, opened 2026-06-20
auth: add support for Enterprise-Managed Authorization (ID-JAG flow)


decko:
## Summary

The MCP spec has a new official extension — **Enterprise-Managed Authorization** (`io.modelcontextprotocol/enterprise-managed-authorization`) — that centralizes MCP server access control through an organization's IdP (Okta, Azure AD, corporate SSO). It replaces the per-user, per-server OAuth2 consent dance with a delegated token exchange flow.

References:
- https://blog.modelcontextprotocol.io/posts/enterprise-managed-auth/
- https://modelcontextprotocol.io/extensions/auth/enterprise-managed-authorization
- https://github.com/modelcontextprotocol/ext-auth

---

## How the flow works

```
Client → (SSO) → Enterprise IdP → ID Token
Client → IdP   → exchanges ID Token for ID-JAG (Identity Assertion JWT Authorization Grant)
Client → MCP Authorization Server → exchanges ID-JAG for MCP Access Token
Client → MCP Resource Server → calls API with Access Token
```

Key difference from current `oauth2` provider: the client **never redirects the user to the MCP server's authorization endpoint**. Access decisions are made entirely at the IdP level (group membership, roles, conditional access policies).

---

### 1. Capability advertisement (`pkg/handler/`)
Declare extension support in the `initialize` request:
```json
{
  "capabilities": {
    "extensions": {
      "io.modelcontextprotocol/enterprise-managed-authorization": {}
    }
  }
}
```

### 2. New auth provider (`internal/auth/enterprise_idp.go`)
Implement the `Provider` interface for the ID-JAG exchange flow:
- Accept IdP configuration: issuer URL, JWKS endpoint, token exchange endpoint, client credentials
- Store the Identity Assertion (ID Token or SAML assertion) from the initial SSO login
- On `Authenticate()`: exchange ID Token → ID-JAG → MCP access token; cache and refresh as needed
- Validate ID-JAG JWT signatures against the IdP's JWKS endpoint before forwarding

### 3. Config schema (`internal/config/`)
New auth variant, e.g.:
```yaml
auth:
  type: enterprise_managed
  idp_issuer: https://company.okta.com
  idp_token_endpoint: https://company.okta.com/oauth2/v1/token
  idp_jwks_uri: https://company.okta.com/oauth2/v1/keys
  client_id: ...
  identity_assertion_file: ~/.config/wtmcp/credentials/id_token.json
```

[...]

PR #186, opened 2026-07-27
feat: support streamable HTTP transport for sidecar deployments

### LINKED ISSUE #179: Feature: support HTTP/SSE transport for sidecar and remote deployments
## Problem

wtmcp only supports stdio transport today. This limits it to local, co-process usage (e.g., Claude Code spawning it directly). There's no way to run wtmcp as a standalone service that remote MCP clients connect to over the network.

## Use case: AI workflow sidecar

In AI workflow platforms like Alcove, agent sessions run in isolated containers (skiff pods). Agents need access to external tools (Jira, GitHub, GitLab, etc.) but the orchestrator controls which tools are available. Running wtmcp as a sidecar container in the agent pod would give agents full MCP tool access over localhost — no custom bridge actions needed per tool.

Current workaround: the workflow engine implements individual bridge actions for each external API (e.g., `jira-create-issue`, `jira-search-issues`). This doesn't scale — every new tool capability requires a code change in the orchestrator.

With HTTP/SSE transport, wtmcp becomes a drop-in sidecar: the agent connects to `http://localhost:<port>` and gets access to all configured plugins.

## Proposed solution

Add a `--transport` flag (or config option) supporting:
- `stdio` (default, current behavior)
- `sse` — Server-Sent Events transport
- `streamable-http` — Streamable HTTP transport (MCP spec 2025-03-26)

The `mcp-go` library already provides `mcpserver.NewSSEServer()` and `mcpserver.NewStreamableHTTPServer()`, so the core wiring should be straightforward.

### Minimum viable scope

```
wtmcp serve --transport sse --port 8080
wtmcp serve --transport streamable-http --port 8080
```

### Config file alternative

```yaml
server:
  transport: sse      # stdio | sse | streamable-http
  port: 8080
  host: 0.0.0.0       # default: localhost
```

## Security considerations

- Stdio is inherently local. HTTP transports expose wtmcp to network access — even `localhost` in a shared pod has a broader blast radius.
- Consider: auth token requirement for HTTP transports, bind address restriction (default localhost), optional TLS.
- In the sidecar use case, the pod network namespace provides isolation, so localhost binding without auth is likely sufficient.

## Context

This came up while designing the Agentic SDLC pipeline for the Hosted Pulp team. The Task Elaborator workflow needs to create multiple Jira issues dynamically. Rather than adding batch bridge actions to the orchestrator for every tool, running wtmcp as a sidecar would give agents direct access to the full Jira toolkit (50+ tools) plus GitHub, GitLab, Confluence, etc.

decko:
## Summary

Closes #179

Adds `streamable-http` transport so wtmcp can run as a standalone HTTP server for sidecar deployments in AI workflow pods. Agents connect over `http://localhost:<port>/mcp` and get access to all configured plugins.

- **Config:** `ServerConfig` with `transport`, `host`, `port` fields (defaults: `stdio`/`localhost`/`8080`)
- **Transport package:** `internal/transport/` with `ListenAndServe()` supporting stdio and streamable-http
- **CLI:** `--transport`, `--host`, `--port` flags on `serve` command; root command forces stdio for backward compatibility
- **Health endpoint:** `/healthz` returns `{"status":"ok"}` for Kubernetes probes
- **Shutdown sequencing:** HTTP drain completes before plugins shut down
- **Reload safety:** Per-plugin RWMutex serializes reload against in-flight tool calls
- **Control info:** Listen URL written to `mcp.info` for external tooling discovery

### Usage

```bash
# Config file
server:
  transport: streamable-http
  host: localhost
  port: 8080

# CLI flags
wtmcp serve --transport streamable-http --port 9090

# Backward compatible — bare wtmcp always uses stdio
wtmcp
```

### Security

- Defaults to `localhost` binding (not exposed to network)
- mcp-go DNS rebinding protection enabled by default
- No auth in v1 — #116 covers Enterprise-Managed Authorization for future HTTP transport security

### mcp-go options wired

- `WithSessionIdleTTL(30min)` — prevents session state memory leaks
- `WithHeartbeatInterval(30s)` — keeps SSE connections alive through proxies
- `WithStreamableHTTPLogger` — routes transport errors to wtmcp log file

## Test plan

- [ ] Config: defaults, YAML parsing, partial overrides, validation (transport, port range, empty host)
- [ ] Transport: stdio cancellation, HTTP start + healthz, port-in-use error propagation, MCP JSON-RPC endpoint, unsupported transport
- [ ] Concurrent sessions: 5 parallel MCP clients with session isolation
- [ ] Reload lock serialization: write lock blocks until read locks release
- [ ] CLI: flag overrides, root command forces stdio
- [ ] Full suite: `make build && make test && make lint` — 30 packages, 0 failures, 0 lint issues

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/cmd/wtmcp/main.go
+++ b/cmd/wtmcp/main.go
@@ -321,7 +321,9 @@ func run(forceStdio bool) error {
 	}
 
 	// Non-plugin cleanup runs on context cancellation.
+	cleanupDone := make(chan struct{})
 	go func() {
+		defer close(cleanupDone)
 		<-ctx.Done()
 		controlWatcher.Stop()
 		cacheStore.Close() //nolint:errcheck,gosec // best-effort on shutdown
@@ -331,6 +333,8 @@ func run(forceStdio bool) error {
 
 	logger := slog.New(slog.NewTextHandler(log.Writer(), &slog.HandlerOptions{Level: slog.LevelInfo}))
 	err = transport.ListenAndServe(ctx, srv, &cfg.Server, logger, os.Stdin, os.Stdout)
+
+	<-cleanupDone // ensure no reload in progress
 
 	// Sequential shutdown: transport drained, now safe to tear down.
 	log.Println("shutting down plugins...")
--- a/cmd/wtmcp/main.go
+++ b/cmd/wtmcp/main.go
@@ -293,23 +293,6 @@ func run(forceStdio bool) error {
 		log.Printf("all plugins loaded (%d)", len(mgr.LoadedPlugins()))
 	}()
 
-	// Start control directory watcher for external reload triggers
-	controlWatcher := server.NewControlWatcher(wd, srv, mgr, cfg, index, collector, auditor, pluginRL, framer, toolOwners)
-	if err := controlWatcher.Start(); err != nil {
-		log.Printf("control watcher disabled: %v", err)
-	}
-
-	// Non-plugin cleanup runs on context cancellation.
-	go func() {
-		<-ctx.Done()
-		controlWatcher.Stop()
-		cacheStore.Close() //nolint:errcheck,gosec // best-effort on shutdown
-		if collector != nil {
-			collector.Close()
-		}
-		auditor.Close() //nolint:errcheck,gosec // best-effort on shutdown
-	}()
-
 	// Apply CLI flag overrides for serve command.
 	if !forceStdio {
 		if transportFlag != "" {
@@ -324,6 +307,25 @@ func run(forceStdio bool) error {
 	} else {
 		cfg.Server.Transport = config.TransportStdio
 	}
+
+	// Start control directory watcher for external reload triggers.
+	// Must come after CLI flag overrides so listenURL reflects the actual transport.
+	listenURL := transport.ListenURL(&cfg.Server)
+	controlWatcher := server.NewControlWatcher(wd, srv, mgr, cfg, index, collector, auditor, pluginRL, framer, toolOwners, listenURL)
+	if err := controlWatcher.Start(); err != nil {
+		log.Printf("control watcher disabled: %v", err)
+	}
+
+	// Non-plugin cleanup runs on context cancellation.
+	go func() {
+		<-ctx.Done()
+		controlWatcher.Stop()
+		cacheStore.Close() //nolint:errcheck,gosec // best-effort on shutdown
+		if collector != nil {
+			collector.Close()
+		}
+		auditor.Close() //nolint:errcheck,gosec // best-effort on shutdown
+	}()
 
 	log.Printf("wtmcp %s starting (workdir: %s, transport: %s)", Version, wd, cfg.Server.Transport)
 
--- a/cmd/wtmcp/main.go
+++ b/cmd/wtmcp/main.go
@@ -308,6 +308,10 @@ func run(forceStdio bool) error {
 		cfg.Server.Transport = config.TransportStdio
 	}
 
+	if err := cfg.Server.Validate(); err != nil {
+		return fmt.Errorf("server config: %w", err)
+	}
+
 	// Start control directory watcher for external reload triggers.
 	// Must come after CLI flag overrides so listenURL reflects the actual transport.
 	listenURL := transport.ListenURL(&cfg.Server)
@@ -321,10 +325,6 @@ func run(forceStdio bool) error {
 		<-ctx.Done()
 		controlWatcher.Stop()
 		cacheStore.Close() //nolint:errcheck,gosec // best-effort on shutdown
-		if collector != nil {
-			collector.Close()
-		}
-		auditor.Close() //nolint:errcheck,gosec // best-effort on shutdown
 	}()
 
 	log.Printf("wtmcp %s starting (workdir: %s, transport: %s)", Version, wd, cfg.Server.Transport)
@@ -332,11 +332,14 @@ func run(forceStdio bool) error {
 	logger := slog.New(slog.NewTextHandler(log.Writer(), &slog.HandlerOptions{Level: slog.LevelInfo}))
 	err = transport.ListenAndServe(ctx, srv, &cfg.Server, logger, os.Stdin, os.Stdout)
 
-	// Plugins shut down after transport stops — ensures in-flight
-	// HTTP requests complete before plugin processes are killed.
+	// Sequential shutdown: transport drained, now safe to tear down.
 	log.Println("shutting down plugins...")
 	mgr.WaitLoaded()
 	mgr.ShutdownAll(context.Background())
+	if collector != nil {
+		collector.Close()
+	}
+	auditor.Close() //nolint:errcheck,gosec // best-effort on shutdown
 
 	return err
 }
--- a/cmd/wtmcp/main.go
+++ b/cmd/wtmcp/main.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"log/slog"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -13,8 +14,6 @@ import (
 	"syscall"
 
 	"github.com/spf13/cobra"
-
-	mcpserver "github.com/mark3labs/mcp-go/server"
 
 	"github.com/LeGambiArt/wtmcp/internal/audit"
 	"github.com/LeGambiArt/wtmcp/internal/auth"
@@ -27,6 +26,7 @@ import (
 	"github.com/LeGambiArt/wtmcp/internal/sandbox"
 	"github.com/LeGambiArt/wtmcp/internal/server"
 	"github.com/LeGambiArt/wtmcp/internal/stats"
+	"github.com/LeGambiArt/wtmcp/internal/transport"
 )
 
 // Version and BuildDate are set via ldflags at build time.
@@ -40,6 +40,10 @@ var (
 	configPath string
 	workdir    string
 	readOnly   bool
+
+	transportFlag string
+	hostFlag      string
+	portFlag      int
 )
 
 var rootCmd = &cobra.Command{
@@ -50,7 +54,7 @@ var rootCmd = &cobra.Command{
 	SilenceErrors: true,
 	// Default action: run the server (backward compatible with MCP clients).
 	RunE: func(_ *cobra.Command, _ []string) error {
-		return run()
+		return run(true)
 	},
 }
 
@@ -58,7 +62,7 @@ var serveCmd = &cobra.Command{
 	Use:   "serve",
 	Short: "Start MCP server (default)",
 	RunE: func(_ *cobra.Command, _ []string) error {
-		return run()
+		return run(false)
 	},
 }
 
@@ -99,6 +103,10 @@ func init() {
 		fmt.Sprintf("wtmcp %s (built %s)\n", Version, BuildDate))
 	rootCmd.DisableAutoGenTag = true
 
+	serveCmd.Flags().StringVar(&transportFlag, "transport", "", "Transport: stdio, streamable-http")
+	serveCmd.Flags().StringVar(&hostFlag, "host", "", "Bind address (default: localhost)")
+	serveCmd.Flags().IntVar(&portFlag, "port", 0, "Listen port (default: 8080)")
+
 	rootCmd.AddCommand(serveCmd, checkCmd, versionCmd)
 }
 
@@ -109,7 +117,7 @@ func main() {
 	}
 }
 
-func run() error {
+func run(forceStdio bool) error {
 	// Capture the caller's CWD for file I/O before anything changes it.
 	sessionDir, err := os.Getwd()
 	if err != nil || !filepath.IsAbs(sessionDir) {
@@ -291,6 +299,7 @@ func run() error {
 		log.Printf("control watcher disabled: %v", err)
 	}
 
+	// Non-plugin cleanup runs on context cancellation.
 	go func() {
 		<-ctx.Done()
 		controlWatcher.Stop()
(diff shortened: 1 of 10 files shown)


internal/transport/transport.go
```diff
@@ -0,0 +1,106 @@
...
+	select {
+	case err := <-errCh:
+		return fmt.Errorf("http server: %w", err)
+	case <-ctx.Done():
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+		if err := httpSrv.Shutdown(shutdownCtx); err != nil {
+			logger.Error("http shutdown error", "err", err)
+		}
+		return nil
```
sergio-correia, line 98: When both `errCh` and `ctx.Done()` are ready, Go's `select` picks one at random. If `ctx.Done()` wins, we return nil here and the startup error is silently lost. Concrete case: port is in use, SIGTERM arrives in the same tick, process exits 0 with no indication the server never bound.

Would be good to drain `errCh` after shutdown:

```go
// After httpSrv.Shutdown:
select {
case err := <-errCh:
    if err != nil && !errors.Is(err, http.ErrServerClosed) {
        return fmt.Errorf("http server: %w", err)
    }
default:
}
return nil
```

Also worth considering: if `Shutdown` returns `context.DeadlineExceeded` (drain timeout exceeded), remaining connections stay open. Calling `httpSrv.Close()` as a fallback would deterministically terminate them before plugin shutdown.

internal/transport/transport.go
```diff
@@ -0,0 +1,106 @@
...
+	mux.Handle("/mcp", httpSrv)
+	mux.HandleFunc("/healthz", handleHealthz)
+
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- httpSrv.Start(addr)
+	}()
+
+	listenURL := fmt.Sprintf("http://%s/mcp", addr)
+	logger.Info("wtmcp starting", "transport", "streamable-http", "addr", listenURL)
```
sergio-correia, line 83: With no auth in v1, binding to a non-loopback address exposes every plugin tool to the network. In a shared K8s cluster, `host: 0.0.0.0` is the obvious thing to try when localhost doesn't work from another container, and there's no indication of the security implication.

Would be good to log a warning here when the host isn't loopback:

```go
if cfg.Host != "localhost" && cfg.Host != "127.0.0.1" && cfg.Host != "::1" {
    logger.Warn("binding to non-loopback address with no authentication",
        "host", cfg.Host, "port", cfg.Port)
}
```

cmd/wtmcp/main.go
```diff
@@ -285,37 +293,54 @@ func run() error {
...
+	logger := slog.New(slog.NewTextHandler(log.Writer(), &slog.HandlerOptions{Level: slog.LevelInfo}))
+	err = transport.ListenAndServe(ctx, srv, &cfg.Server, logger, os.Stdin, os.Stdout)
 
-	// Signal-initiated shutdown is not an error.
-	if err != nil && ctx.Err() != nil {
-		return nil
+	// Sequential shutdown: transport drained, now safe to tear down.
+	log.Println("shutting down plugins...")
+	mgr.WaitLoaded()
+	mgr.ShutdownAll(context.Background())
```
sergio-correia, line 342: There's a race between this background goroutine and `ShutdownAll` below. When SIGTERM arrives during an in-progress reload, `controlWatcher.Stop()` waits for it to finish (which may be creating a new handle). Meanwhile, `ListenAndServe` returns after HTTP drain and `ShutdownAll` scans the handles map. If the reload is between `Unload` (old handle deleted) and `Load` (new handle inserted), `ShutdownAll` misses the new process.

Fix: wait for this goroutine before `ShutdownAll`:

```go
cleanupDone := make(chan struct{})
go func() {
    defer close(cleanupDone)
    <-ctx.Done()
    controlWatcher.Stop()
    cacheStore.Close()
}()

err = transport.ListenAndServe(...)

<-cleanupDone  // ensure no reload in progress
```

internal/server/server.go
```diff
@@ -687,6 +687,18 @@ func ReloadPlugin(ctx context.Context, srv *mcpserver.MCPServer, mgr *plugin.Man
...
 
+	// Acquire the reload write lock on the current handle to wait for
+	// in-flight tool calls to complete before tearing it down. The defer
+	// unlocks the OLD handle after Reload replaces it — this is safe
+	// because (a) the ControlWatcher serializes reload commands, and
+	// (b) srv.AddTool atomically replaces tool handlers, so new calls
+	// route to the new handle (with its own fresh mutex) only after
+	// it is fully initialized.
+	if handle := mgr.Handle(name); handle != nil {
+		handle.ReloadLock()
```
sergio-correia: `mgr.Reload()` calls `m.WaitLoaded()` internally (manager.go:832). Since the write lock is acquired here *before* `Reload()`, any early reload triggered during initial startup (e.g., a `reload-pluginA` control file while other plugins are still loading) holds this write lock for the entire remaining startup duration. All tool calls to this plugin block at `reloadMu.RLock()` until loading finishes.

Moving `WaitLoaded()` before the lock acquisition fixes it:

```go
mgr.WaitLoaded()

if handle := mgr.Handle(name); handle != nil {
    handle.ReloadLock()
    defer handle.ReloadUnlock()
}
```

The redundant `WaitLoaded()` inside `Reload()` becomes a no-op.

internal/server/server.go
```diff
@@ -687,6 +687,18 @@ func ReloadPlugin(ctx context.Context, srv *mcpserver.MCPServer, mgr *plugin.Man
 		collector.RemovePluginResources(name)
 	}
 
+	// Acquire the reload write lock on the current handle to wait for
+	// in-flight tool calls to complete before tearing it down. The defer
+	// unlocks the OLD handle after Reload replaces it — this is safe
+	// because (a) the ControlWatcher serializes reload commands, and
```
sergio-correia: This assumption holds for control file reloads (ControlWatcher processes them sequentially), but with HTTP transport, two MCP sessions can call `plugin_reload` concurrently. Both calls resolve `mgr.Handle(name)` to the same handle, so the second caller locks a stale handle after the first reload replaces it. In-flight calls on the new handle aren't drained by the second reload.

`mgr.reloadMu` serializes the actual reloads so no data corruption happens, but in-flight calls get "plugin exited" errors unnecessarily.

Consider moving the per-handle lock acquisition into `Manager.Reload()`, after `m.reloadMu.Lock()`, where it always sees the current handle:

```go
if handle := m.handles[name]; handle != nil {
    handle.ReloadLock()
    defer handle.ReloadUnlock()
}
```

internal/transport/transport_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+	wg.Wait()
+	close(errors)
+
+	for err := range errors {
+		t.Error(err)
+	}
+}
+
+func TestReloadLockSerializesAgainstCalls(t *testing.T) {
```
sergio-correia: This test verifies that `sync.RWMutex` serializes read/write locks, which is a property guaranteed by Go's standard library. It doesn't exercise the actual `Handle.CallTool()` or `Handle.ReloadLock()` methods.

If someone removes `h.reloadMu.RLock()` from `Handle.CallTool()` (dispatch.go:126), this test still passes and the feature's invariant is silently broken.

Would be better to test against the actual `Handle` type in `internal/plugin/dispatch_test.go`: start a slow tool call, attempt `ReloadLock()` concurrently, verify it blocks until the call completes.

sergio-correia (commented):
The core safety claim ("HTTP drain completes before plugins shut down") exists as sequential code ordering in main.go with no test coverage. A refactor that moves `ShutdownAll` into the context cancellation goroutine would break the drain guarantee with no test failure. An integration test that starts a slow tool call over HTTP, cancels the context mid-call, and verifies the response completes would protect this invariant.

internal/transport/transport.go
decko: Fixed in 326fe0b. After `Shutdown()`, we now drain `errCh` to catch startup errors masked by simultaneous context cancellation:

```go
select {
case err := <-errCh:
    if err != nil && !errors.Is(err, http.ErrServerClosed) {
        return fmt.Errorf("http server: %w", err)
    }
default:
}
```

On the `Close()` fallback: mcp-go's `StreamableHTTPServer` doesn't expose `Close()`, so we log the error and let the process exit. The shutdown timeout (5s) + process exit covers the case.

internal/transport/transport.go
decko: Fixed in 326fe0b. Added warning before the listen log:

```go
if cfg.Host != "localhost" && cfg.Host != "127.0.0.1" && cfg.Host != "::1" {
    logger.Warn("binding to non-loopback address with no authentication",
        "host", cfg.Host, "port", cfg.Port)
}
```

cmd/wtmcp/main.go
decko: Fixed in f43361d. The cleanup goroutine now signals completion via a channel, and `ShutdownAll` waits for it:

```go
cleanupDone := make(chan struct{})
go func() {
    defer close(cleanupDone)
    <-ctx.Done()
    controlWatcher.Stop()
    cacheStore.Close()
}()

err = transport.ListenAndServe(...)
<-cleanupDone // ensure no reload in progress
mgr.ShutdownAll(context.Background())
```

internal/server/server.go
decko: Fixed in 3aeac84. Moved the handle-level lock acquisition into `Manager.Reload()`, after `m.reloadMu.Lock()` and after `WaitLoaded()`. This addresses both this finding and the concurrent reload finding below — the lock always sees the current handle since it's inside the manager's serialization mutex.

internal/server/server.go
decko: Fixed in the same commit (3aeac84). The lock acquisition moved into `Manager.Reload()`, after `m.reloadMu.Lock()`:

```go
m.reloadMu.Lock()
defer m.reloadMu.Unlock()

m.handlesMu.Lock()
handle := m.handles[name]
m.handlesMu.Unlock()
if handle != nil {
    handle.ReloadLock()
    defer handle.ReloadUnlock()
}
```

This guarantees the lock always targets the current handle, regardless of concurrent reloads from HTTP sessions. The server-side `ReloadPlugin` no longer acquires the handle lock itself.

internal/transport/transport_test.go
decko: Renamed to `TestRWMutexReloadPattern` in f43361d with a comment clarifying it validates the locking pattern, not Handle integration:

```go
// TestRWMutexReloadPattern validates the locking pattern used by
// Handle.reloadMu: read locks (tool calls) run concurrently, but a
// write lock (reload) blocks until all reads complete. This tests the
// pattern, not the Handle integration — a Handle-level test requires
// a running plugin process. See dispatch.go:126 for the RLock site.
```

A Handle-level integration test would require a running plugin subprocess. We added `TestListenAndServeHTTPDrainBeforeShutdown` instead, which exercises the full HTTP → MCP → tool call → shutdown drain path end-to-end.

decko:
$(cat <<'EOF'
All 7 findings addressed:

| # | Finding | Fix | Commit |
|---|---------|-----|--------|
| 1 | `select` race: errCh vs ctx.Done() | Drain errCh after Shutdown | `326fe0b` |
| 2 | Non-loopback binding with no auth | Log warning when host is not loopback | `326fe0b` |
| 3 | Cleanup goroutine race with ShutdownAll | Wait for cleanup via channel before ShutdownAll | `f43361d` |
| 4 | WaitLoaded() inside write lock | Moved lock into Manager.Reload(), after WaitLoaded | `3aeac84` |
| 5 | Concurrent plugin_reload via HTTP | Same: lock acquired inside Manager.Reload() after reloadMu | `3aeac84` |
| 6 | No integration test for drain guarantee | Added `TestListenAndServeHTTPDrainBeforeShutdown` | `f43361d` |
| 7 | Reload test tests stdlib, not Handle | Renamed + clarified scope; drain test covers the full path | `f43361d` |

Branch rebased onto latest main (includes mcp-go v0.57.0 bump). All 30 packages pass, 0 lint issues.
EOF
)

internal/transport/transport.go
```diff
@@ -0,0 +1,120 @@
...
+	case <-ctx.Done():
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+		if err := httpSrv.Shutdown(shutdownCtx); err != nil {
+			logger.Error("http shutdown error, forcing close", "err", err)
+		}
+		// Drain startup error if Start() failed in the same tick as ctx cancellation.
+		select {
+		case err := <-errCh:
+			if err != nil && !errors.Is(err, http.ErrServerClosed) {
```
sergio-correia: After `Shutdown()` returns, `Start()` is guaranteed to have already returned (the listener was closed during `Shutdown()`). So a blocking drain here is safe and would be cleaner:

```go
if err := <-errCh; err != nil && !errors.Is(err, http.ErrServerClosed) {
    return fmt.Errorf("http server: %w", err)
}
return nil
```

The non-blocking `select` with `default` can miss the error if the Start goroutine hasn't been scheduled yet. Unlikely in practice, but unnecessary when we know `Start()` has returned.

internal/plugin/manager.go
```diff
@@ -838,6 +838,16 @@ func (m *Manager) Reload(ctx context.Context, name string) error {
 	m.reloadMu.Lock()
 	defer m.reloadMu.Unlock()
 
+	// Lock the current handle to wait for in-flight tool calls.
+	// Must be inside reloadMu to always see the current handle.
+	m.handlesMu.Lock()
```
sergio-correia, line 843: After a reload completes and `ReloadUnlock()` fires, a tool call that was blocked at `reloadMu.RLock()` (dispatch.go:126) since before the reload started will proceed with the now-dead handle, producing a `send_failed: broken pipe` error. The window is narrow (between map lookup and RLock), the error is non-destructive, and a retry succeeds.

Consider checking `h.process.State()` after acquiring the read lock in `CallTool` and returning a `plugin_reloading` error instead of the raw broken-pipe. Much clearer for the operator.

internal/transport/transport.go
```diff
@@ -0,0 +1,120 @@
...
+// in control info files. Returns "stdio" for stdio transport.
+func ListenURL(cfg *config.ServerConfig) string {
+	if cfg.Transport == config.TransportStdio {
+		return "stdio"
+	}
+	addr := net.JoinHostPort(cfg.Host, fmt.Sprintf("%d", cfg.Port))
+	return fmt.Sprintf("http://%s/mcp", addr)
+}
+
+func handleHealthz(w http.ResponseWriter, _ *http.Request) {
```
sergio-correia, line 116: `/healthz` always returns 200 regardless of plugin loading state. For K8s readiness probes, this means the pod is marked ready before tools are available. Consider adding a `/readyz` endpoint that checks loading state, or documenting that `/healthz` is liveness-only.

sergio-correia (commented):
Not in this diff, but triggered by the `cleanupDone` fix: `Stop()` at `internal/server/control.go:87` blocks on `<-w.done`, which is only closed by `pollLoop()` (line 96). If `Start()` returns an error before launching `pollLoop()`, `Stop()` deadlocks forever.

With the `cleanupDone` barrier from f43361d, this now blocks the entire shutdown path: `<-cleanupDone` at main.go:341 never returns, so `ShutdownAll`, `collector.Close`, and `auditor.Close` never run. Process hangs until SIGKILL.

Scenario: workdir is read-only or disk full, `Start()` fails at `os.MkdirAll`, server runs normally, SIGTERM arrives, deadlock.

Please guard `Stop()` with a started flag (e.g., `atomic.Bool`) so it's safe to call when `Start()` never succeeded.

internal/transport/transport.go
decko: Fixed in 460c408. Switched to a blocking drain since Start() is guaranteed to have returned after Shutdown() closes the listener:

```go
if err := <-errCh; err != nil && !errors.Is(err, http.ErrServerClosed) {
    return fmt.Errorf("http server: %w", err)
}
```

internal/plugin/manager.go
decko: Fixed in 460c408. CallTool now checks for StateStopping after acquiring the read lock and returns a clear `plugin_reloaded` error with retry guidance instead of hitting a broken pipe.

internal/transport/transport.go
decko: Documented `/healthz` as liveness-only in 460c408. A `/readyz` endpoint that checks plugin loading state is deferred — it requires passing plugin manager state into the transport package, which is scope expansion for this PR.

decko:
Re: `controlWatcher.Stop()` deadlock when `Start()` fails — fixed in 460c408. Added `atomic.Bool` started flag; `Stop()` returns early if `Start()` never succeeded. The `cleanupDone` barrier no longer deadlocks on the `Start()` failure path.

sergio-correia (approved):
LGTM, thanks!

PR #191, opened 2026-08-04
feat(gitlab): add target_project_id to gitlab_create_merge_request

Cropi:
The tool only supported same-project MRs. The typical fork-to-upstream workflow (push to your fork, open MR against the upstream) didn't work because there was no way to specify the target project.

This adds an optional `target_project_id` param. Pass your fork as `project_id` and the upstream as `target_project_id` -- either a numeric ID or a path like `redhat/centos-stream/rpms/audit`, the tool resolves it either way.

A few things worth noting:

- Paths are validated (regex + 255 char cap) before hitting the API
- Negative/zero numeric IDs are rejected early
- Lookup failures return a generic error -- no 404 vs 403 distinction to avoid leaking whether a project exists
- If you pass a path, one live GET happens even in dry-run mode to resolve it to a numeric ID; this is documented in the tool description
- The resolved `target_project_id` shows up in both the dry-run preview and the live response

Tested end-to-end against a real GitLab fork (CentOS Stream audit package).

### DIFF
--- a/plugins/gitlab/plugin.yaml
+++ b/plugins/gitlab/plugin.yaml
@@ -389,6 +389,9 @@ tools:
         type: string
         required: true
         description: "Branch to merge into"
+      target_project_id:
+        type: string
+        description: "Target project ID or path for cross-project (fork-to-upstream) MRs. When set, project_id is the fork and this is the upstream project. A path value is resolved to a numeric ID via a live API call, even in dry-run mode."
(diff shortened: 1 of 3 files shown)


Cropi:
The pre-commit failures are preexisting issues.

sergio-correia (approved):
LGTM, thanks!

PR #200, opened 2026-08-17
sandbox: add cgroup trampoline and AllowExec for Python plugins

sergio-correia:
- Add cgroup trampoline that re-execs through systemd-run --user --scope when the user inherits a root-owned cgroup (common in SSH/su sessions), so sandbox resource limits work without manual systemd configuration
- Set AllowExec on sandbox profiles for Python plugin handlers so Landlock permits execve on the interpreter
- Set CgroupBestEffort so the sandbox degrades gracefully when cgroup controllers aren't delegated instead of hard-failing
- Bump arapuca to 0.2.7 and go-arapuca to 0.2.4 (provide the required FFI)

### DIFF
--- a/cmd/wtmcp/cgroup_trampoline_linux.go
+++ b/cmd/wtmcp/cgroup_trampoline_linux.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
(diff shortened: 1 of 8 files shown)


rjeffman (approved):
LGTM.

PR #211, opened 2026-08-19
credentials: resolve relative credential paths before vault check

sergio-correia:
The credential integration commit (6fb9207) changed decryptCredFile's error handler from returning the original path to returning an empty string. This broke OAuth2 plugins that specify credentials_file as a relative path (e.g., "client-credentials.json" in google-drive's plugin.yaml): os.Stat fails because the path is relative to CWD, decryptCredFile returns "", and NewOAuth2Provider rejects the empty path with "credential file path is required". The cascade disables auth entirely — the proxy makes unauthenticated requests, and Google returns 403.

Fix by resolving relative credential paths against the credential group directory before passing to decryptCredFile, so os.Stat can find the file. Apply resolveCredPath to TokenFile as well, since refresh_token auth type calls loadRefreshToken which uses os.Stat directly without resolveCredentialPath. Affects CredentialsFile, TokenFile, and PrivateKeyFile in both variant and single-type auth config paths.


decko (commented):
## Review (Go + MCP specialists)

Relative Google OAuth paths are fixed. Merge after a credential-dir jail **before** any Stat/Read.

**Critical**
- `resolveCredPath` has no stay-inside-`credDir` check. Vault Stat/Read happens before the oauth2 jail. `TokenFile` / `PrivateKeyFile` never hit that jail (`loadRefreshToken` / `loadPrivateKeyFile`).
- Vault-encrypted `client-credentials.json`: decrypt returns `/proc/self/fd/N`, then oauth2 `resolveCredentialPath` / `RejectSymlink` reject it → `resolveAuth` nil → unauthenticated MCP calls / Google 403.

**Important**
- Test is oauth2 happy-path only. Add traversal, variants, `refresh_token`, `github_app`, vault.

**Minor**
- `decryptCredFile` maps every error to `""` → misleading `credential file path is required`.

internal/plugin/manager.go
```diff
@@ -1520,6 +1520,16 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
+	resolveCredPath := func(path string) string {
+		if path == "" || filepath.IsAbs(path) {
+			return path
+		}
+		if credDir != "" {
+			return filepath.Join(credDir, path)
+		}
+		return path
+	}
+
```
decko, line 1539: **Critical:** `filepath.Join` Cleans `..` with no stay-inside-`credDir` check. `decryptCredentialIfVault` then Stats/Reads **before** oauth2's jail.

A plugin.yaml path of `../../../.ssh/id_rsa` or `../jira/...` is read in the unsandboxed core. Reuse `resolveCredentialPath` (Clean + prefix + `EvalSymlinks`) **before** any Stat/Read.

internal/plugin/manager.go
```diff
@@ -1520,6 +1520,16 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
 	decryptCredFile := func(path string) string {
 		if path == "" {
 			return ""
```
decko, line 1542: **Minor:** Every error becomes `""`. Missing file logs "decrypt failed, skipping" then oauth2 reports "credential file path is required". Return the jailed path on `os.IsNotExist`, or plumb the Stat error.

internal/plugin/manager.go
```diff
@@ -1574,15 +1584,15 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
 				Password:        resolve(v.Password),
 				SPN:             resolve(v.SPN),
 				Scopes:          v.Scopes,
-				CredentialsFile: decryptCredFile(resolve(v.CredentialsFile)),
-				TokenFile:       resolve(v.TokenFile),
+				CredentialsFile: decryptCredFile(resolveCredPath(resolve(v.CredentialsFile))),
```
decko, line 1597: **Critical (MCP):** After this join, a vaulted `client-credentials.json` decrypts to `/proc/self/fd/N`. `NewOAuth2Provider` then jails that memfd path (and `RejectSymlink` would reject it) → `resolveAuth` nil → Google 403.

Jail the original group path, decrypt, then load the memfd without a second jail/symlink check. Same call site exists on the single-type path below.

internal/plugin/manager.go
```diff
@@ -1574,15 +1584,15 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
+				TokenFile:       resolveCredPath(resolve(v.TokenFile)),
```
decko, line 1598: **Critical:** `TokenFile` is resolved but never jailed. `loadRefreshToken` uses this path as-is. Same for `PrivateKeyFile` → `loadPrivateKeyFile` (line 1595).

Jail here too; oauth2's later jail does not cover `refresh_token` / `github_app`.

internal/plugin/manager_test.go
```diff
@@ -824,6 +824,67 @@ func createPluginWithManifest(t *testing.T, parentDir, name, manifestYAML string
...
+		t.Fatal("resolveAuth returned nil; relative credential paths should resolve against credential_group directory")
+	}
+	if provider.Name() != "oauth2" {
+		t.Errorf("provider name = %q, want %q", provider.Name(), "oauth2")
+	}
+	if !provider.Available() {
+		t.Error("provider should be available (token file exists)")
+	}
+}
+
```
decko, line 887: **Important:** Happy-path oauth2 only. Would still pass if `TokenFile` resolve were dropped (oauth2 joins relative `token_file` itself).

Add: `..` escape (must fail), variants, `refresh_token` `TokenFile`, `github_app` `PrivateKeyFile`, vault memfd.

internal/plugin/manager.go
sergio-correia: Done. `resolveCredPath` now applies `filepath.Clean` on the joined path and rejects it if it escapes `credDir` via `strings.HasPrefix`. Added `TestResolveAuthRejectsPathTraversal` that confirms `../stolen.json` is rejected.

internal/plugin/manager.go
sergio-correia: Done. `decryptCredFile` now returns the resolved path on `os.IsNotExist` so downstream gets a clear "no such file" error instead of the confusing "credential file path is required".

internal/plugin/manager.go
sergio-correia: Done. Added `isFdPath` exemption in `resolveCredentialPath` — matches `/dev/fd/N` (macOS) and `/proc/{self,<pid>}/fd/N` (Linux, after `EvalSymlinks` resolves the `/proc/self` symlink). Fd component is validated as numeric with no trailing segments.

internal/plugin/manager.go
sergio-correia: Done. Both are now wrapped in `resolveCredPath` in both variant and single-type paths — same containment check as `CredentialsFile`.

internal/plugin/manager_test.go
sergio-correia: Added `TestResolveAuthRejectsPathTraversal` (traversal rejection) and expanded `TestIsFdPath` with negative cases (`/proc/self/fdx/3`, `/proc/self/fd/abc`, `/proc/self/fd/3/extra`, empty fd, etc.). Variant path and `PrivateKeyFile` tests skipped — shared closure, identical call pattern, low risk of divergence.

internal/plugin/manager.go
```diff
@@ -1520,12 +1520,32 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
 	resolve := func(s string) string { return config.ResolveVars(s, vars) }
 
+	resolveCredPath := func(path string) string {
+		if path == "" || filepath.IsAbs(path) {
+			return path
+		}
+		if credDir == "" {
+			return path
+		}
+		joined := filepath.Clean(filepath.Join(credDir, path))
```
decko, line 1530: **[Security/Go — Major]** `resolveCredPath` uses `filepath.Clean` but not `filepath.EvalSymlinks`. A symlink inside `credDir` (e.g. `credDir/link -> /etc/`) passes the `HasPrefix` check because the joined path lexically starts with `cleanDir`. The existing `resolveCredentialPath` in `oauth2.go` calls `filepath.EvalSymlinks` before the containment check — this closure should do the same.

internal/plugin/manager.go
```diff
@@ -1520,12 +1520,32 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
+		cleanDir := filepath.Clean(credDir)
+		if joined != cleanDir &&
+			!strings.HasPrefix(joined, cleanDir+string(filepath.Separator)) {
+			log.Printf("[%s] credential path escapes group directory: %s", pluginName, path)
+			return ""
```
decko, line 1535: **[OAuth2/Go — Major]** On path traversal (`../` escape), this returns `""` with only a log line. For `PrivateKeyFile` and `CredentialsFile`, an empty string causes the provider to fall back silently or fail with a confusing downstream error rather than hard-rejecting the manifest. Consider returning an error and aborting provider construction, or at minimum logging at a higher severity so misconfigurations are surfaced immediately.

internal/plugin/manager.go
```diff
@@ -1520,12 +1520,32 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
+		return joined
+	}
+
 	decryptCredFile := func(path string) string {
 		if path == "" {
 			return ""
 		}
 		decrypted, err := m.decryptCredentialIfVault(pluginName, path)
 		if err != nil {
+			if os.IsNotExist(err) {
```
decko, line 1546: **[OAuth2/Security — Major]** Returning `path` on `os.IsNotExist` can silently bypass vault decryption. If a vault-encrypted credential file is missing but a plaintext file with the same path happens to exist on disk, the provider receives the plaintext path and reads it without decryption. For `CredentialsFile` and `PrivateKeyFile` this is a vault-bypass. Also, this re-exposes the original bug for the ENOENT case — the resolved path is returned but it may still be relative if `resolveCredPath` was somehow skipped. At minimum, log this branch so vault-skip events are observable.

internal/auth/oauth2.go
```diff
@@ -396,7 +396,47 @@ func resolveCredentialPath(path, credentialsDir string) (string, error) {
...
+// /proc/self symlink). The fd component must be numeric with no
+// trailing path segments.
+func isFdPath(path string) bool {
+	if after, ok := strings.CutPrefix(path, "/dev/fd/"); ok {
+		return isNumericFd(after)
+	}
+	if !strings.HasPrefix(path, "/proc/") {
+		return false
+	}
+	rest := path[len("/proc/"):]
```
decko, line 420: **[OAuth2/Security — Major]** The PID segment (`rest[:slash]`) is never validated. `isFdPath` accepts `/proc/<anything>/fd/N` — including other processes' fds (e.g. `/proc/1/fd/3`). An attacker-controlled `CredentialsFile` value pointing at init'\''s open fd would pass the containment exemption. Restrict to `self` or `strconv.Itoa(os.Getpid())`:
```go
pid := rest[:slash]
if pid != "self" && pid != strconv.Itoa(os.Getpid()) {
    return false
}
```

decko (commented):
## Specialist Team Review — PR #211

Three specialists (Go, OAuth2, MCP Security) reviewed this. The fix is directionally correct but the new `resolveCredPath` closure in `manager.go` has weaker security guarantees than the existing `resolveCredentialPath` in `oauth2.go`. Four major issues inline; summary below.

**Major issues (inline):**
1. `resolveCredPath` — no `EvalSymlinks`, symlink escape possible
2. `resolveCredPath` — path traversal silently degrades auth instead of failing closed
3. `isFdPath` — exempts any PID, not just the current process
4. `decryptCredFile` — `os.IsNotExist` fallback can silently bypass vault decryption

**Root cause of all four:** `resolveCredPath` in `manager.go` reimplements the containment logic from `resolveCredentialPath` in `oauth2.go` but without `EvalSymlinks` and without the `isFdPath` exemption, creating two divergent security postures. The recommended fix is to extract `resolveCredentialPath` (or a simplified form of it) into a shared helper and use it in both places.

internal/plugin/manager.go
```diff
@@ -1519,17 +1529,98 @@ func (m *Manager) resolveAuth(pluginName string, manifest *Manifest) auth.Provid
...
+	prepareAuthConfig := func(v AuthServiceConfig) (auth.SingleAuthConfig, error) {
+		credentialsInput := resolve(v.CredentialsFile)
+		if err := rejectManifestFDPath(credentialsInput); err != nil {
+			return auth.SingleAuthConfig{}, fmt.Errorf("credentials_file: %w", err)
+		}
+		credentialsFile, err := resolveCredPath(credentialsInput)
+		if err != nil {
+			return auth.SingleAuthConfig{}, fmt.Errorf("credentials_file: %w", err)
+		}
+		credentialsResult, err := decryptCredFile(credentialsFile)
```
decko: **Breaking change for non-oauth2 plugins.** `prepareAuthConfig` resolves and decrypts `credentials_file` for every auth type before the type is known. A `bearer` or `kerberos` plugin with a stale `credentials_file` field in its manifest will hit ENOENT here, cause `prepareAuthConfig` to return an error, and disable auth entirely. Old behavior: `decryptCredFile` logged and returned `""`, the field was passed through unused, and the provider authenticated normally.

Fix: guard credential-file resolution/decryption behind the auth types that actually require it (`oauth2` for `credentials_file`, `github_app` for `private_key_file`), or treat ENOENT as a warning when the auth type won't use the field.

internal/auth/github_app.go
```diff
@@ -266,3 +267,22 @@ func loadPrivateKeyFile(path string) ([]byte, error) {
...
+	if !IsCurrentProcessFDPath(path) {
+		return nil, fmt.Errorf("private key file is not a current-process fd path")
+	}
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	// procfs and /dev/fd expose synthetic 0777 metadata for trusted
+	// securefiles; the descriptor itself is already access-controlled.
+	if !IsCurrentProcessFDPath(path) {
```
decko: Dead code. The outer guard at line 273 already returns an error when `!IsCurrentProcessFDPath(path)`, so execution only reaches here when the path IS an fd path. This condition is always `false` — `config.CheckPermissions` is never called. If the outer guard is later relaxed to a log warning, permission validation will be silently skipped.

Remove the block (permission checks don't apply to fd-backed securefiles), or restructure so the single-type check happens once.

decko (commented):
Two findings from the post-revision specialist review.

**[Critical] `prepareAuthConfig` breaks non-oauth2 plugins with extraneous credential fields**
`prepareAuthConfig` resolves, stats, and decrypts all credential fields before the auth type is checked. A `bearer`, `kerberos`, or `refresh_token` plugin whose manifest has a stale `credentials_file` or `private_key_file` field will hit ENOENT here, return an error, and disable auth entirely. Previously `decryptCredFile` logged and returned `""` on error, so the provider ignored the unused field and authenticated normally. This is a breaking change for any deployed plugin that matches this shape.

**[Minor] Dead `!IsCurrentProcessFDPath` inner block in two SecureFile loaders**
Both `loadOAuth2ConfigFromSecureFile` (oauth2.go) and `loadPrivateKeySecureFile` (github_app.go) gate entry with `if !IsCurrentProcessFDPath(path) { return nil, err }`, then have a second `if !IsCurrentProcessFDPath(path)` block further down that wraps `config.CheckPermissions`. The inner block is unreachable — `CheckPermissions` is never called. If the outer guard is later relaxed to a warning, permission validation will be silently skipped.

internal/auth/oauth2.go
```diff
@@ -328,6 +368,36 @@ func loadOAuth2Config(path string, scopes []string) (*oauth2.Config, error) {
...
+	if !IsCurrentProcessFDPath(path) {
+		return nil, fmt.Errorf("credentials file is not a current-process fd path")
+	}
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	// procfs and /dev/fd expose synthetic 0777 metadata for trusted
+	// securefiles; the descriptor itself is already access-controlled.
+	if !IsCurrentProcessFDPath(path) {
```
decko: Same dead-code pattern as `loadPrivateKeySecureFile`. The outer guard at line 376 ensures the path is an fd path by the time execution reaches here, so `!IsCurrentProcessFDPath(path)` is always `false` and `config.CheckPermissions` is unreachable. Remove the block.

decko (approved):
All findings from three review rounds addressed. LGTM.

PR #212, opened 2026-08-19
Assorted fixes: build layout, error handling, manifest schema, naming

sergio-correia:
- build: Move wtmcpctl output to project root alongside wtmcp, removing the stale bin/ directory; add a Makefile guard for the legacy ./wtmcp data directory that conflicts with the binary
  - server: Propagate json.Marshal errors in tool and search registration instead of silently dropping them
  - manifest: Support arbitrary nesting depth in ParamsSchema (previously capped at two levels)
  - wtmcpctl: Rename MCP server key from what-the-mcp to wtmcp
  - testing-farm: Use ValueError and ApiError instead of bare Exception in error paths

### DIFF
--- a/cmd/wtmcpctl/agent.go
+++ b/cmd/wtmcpctl/agent.go
@@ -12,7 +12,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const mcpServerKey = "what-the-mcp"
+const mcpServerKey = "wtmcp"
+const legacyServerKey = "what-the-mcp"
 
 // agentSpec defines how a specific AI agent stores its MCP server configuration.
 type agentSpec struct {
@@ -207,8 +208,9 @@ func agentEnable(agentName, dir string, readOnly bool) error {
 		serverEntry["type"] = "stdio"
 	}
 
-	// Set the server entry
+	// Set the server entry and remove legacy key if present
 	mcpServers[mcpServerKey] = serverEntry
+	delete(mcpServers, legacyServerKey)
 	config["mcpServers"] = mcpServers
 
 	// Write config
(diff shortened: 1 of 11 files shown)


cmd/wtmcpctl/agent.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const mcpServerKey = "what-the-mcp"
+const mcpServerKey = "wtmcp"
```
decko, line 15: This is a client identity change, not a rename in place. Enable/disable below only use `mcpServerKey`, so leftover `"what-the-mcp"` entries are never removed.

Clients treat `mcpServers` keys as server identity. After `agent enable`, existing configs run two stdio servers (`what-the-mcp` and `wtmcp`) with duplicated tools.

Keep a legacy alias: on enable, write `wtmcp` then `delete(mcpServers, "what-the-mcp")`. On disable (line 256), delete both keys. Add a test that starts from the old key.

decko (commented):
**Blocker: MCP server key rename has no migration.**

`mcpServerKey` changed from `what-the-mcp` to `wtmcp`, but enable/disable only touch the new key. Existing Claude Code / Cursor / Gemini configs keep `what-the-mcp`.

Re-enable registers both servers (duplicate tools). Disable leaves the old one running.

Treat `"what-the-mcp"` as a legacy alias: delete it on enable, delete both on disable, and add a test that starts from the old key.

cmd/wtmcpctl/agent.go
sergio-correia: Done

rjeffman (approved):
LGTM

decko:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Generated with [Claude Code](https://claude.ai/code)

decko (approved):

PR #213, opened 2026-08-21
plugin: resolve relative auth credential paths

walac:
Auth variants (JWT/service-account) that specify a relative CredentialsFile or PrivateKeyFile were passed straight to the vault decryption helper without first anchoring them to the plugin's credentials directory, so they ended up resolved against the process's current working directory instead of the intended per-plugin or per-credential-group location. Join relative paths with credDir before decrypting, matching the resolution already applied to the base credentials directory a few lines above.


sergio-correia:
Isnt this solving the same issue as https://github.com/LeGambiArt/wtmcp/pull/211?

walac:
@sergio-correia oh, I believe so.

PR #217, opened 2026-08-27
github: add commit_files tool for atomic multi-file commits

Koncpa:
New tool `github_commit_files` uses the Git Data API (blobs + trees + commits) to push multiple files in a single atomic commit. Shell scripts (.sh) automatically get executable permission (100755).

This replaces the pattern of calling `create_or_update_file` per file, which creates one commit per file and always uses mode 100644.


sergio-correia (approved):
LGTM, thanks!

PR #222, opened 2026-09-02
gitlab: add commit_files tool for atomic multi-file commits

Koncpa:
Mirror of github_commit_files — uses GitLab Commits API with actions array to push multiple files in a single commit. Shell scripts (.sh) get executable permission via ExecuteFilemode.


sergio-correia (approved):
LGTM, thanks

PR #223, opened 2026-09-08
sandbox: grant exec access to macOS framework Python version dir

vladi-k:
Framework builds of Python (e.g. the python.org macOS installer) relaunch themselves through a bundled Resources/<Name>.app/Contents/ MacOS/<Name> binary outside bin/. pythonReadPaths() only granted the interpreter's own file and its bin/ directory, so that relaunch was denied by the Seatbelt profile (posix_spawn failing with "Undefined error: 0"), causing plugin init to time out for any Python plugin whenever a framework build was first on PATH.

Detect the enclosing <Name>.framework/Versions/<ver> directory and add it to the sandbox's allowed paths so the relaunch succeeds.

Assisted-by: Claude Sonnet 5 (Anthropic)


sergio-correia (approved):
LGTM, thanks!

PR #238, opened 2026-09-16
github: add reply-to-review-comment tool

sergio-correia:
The GitHub plugin could post new inline review comments and conversation-level comments, but could not reply to an existing inline review comment thread — one of the most common PR operations.

The new tool uses the GitHub REST API replies endpoint and follows the established add_comment pattern. The tool description documents the GitHub API constraint that only top-level review comments can be replied to (replies to replies return 404).


rjeffman (approved):
LGTM.

PR #239, opened 2026-09-16
deps: update arapuca to 0.2.8

sergio-correia:
Move the sandbox library and Go bindings to the LeGambiArt repositories while keeping the compatible go-arapuca 0.2.5 release.


rjeffman (approved):
LGTM

